PR #3, opened 2020-03-14
Update .gitignore

a2k-hanlon:
Added descriptions, and add entry to ignore the Drivers folder (HAL and CMSIS libraries)


lailakhanz (changes requested):
See if you can push more changes after this review

lailakhanz (approved):

PR #4, opened 2020-03-14
Fix faulty .gitignore and add C .gitignore template

a2k-hanlon:
Atom doesn't like comments on the same line as entries in a .gitignore.


RootBeer1313 (approved):
Looks fine; it only ignores generated overhead

PR #16, opened 2021-06-05
User/matthewjegan/bms/ceedling

matthewjegan:
Pull ceedling setup and shiftDchStatus fix discovered by unit tests to master branch


components/bms/bms_master/project.yml
```diff
@@ -0,0 +1,113 @@
+---
+
+# Notes:
+# Sample project C code is not presently written to produce a release artifact.
```
a2k-hanlon: Do we need this template comment still?

a2k-hanlon (changes requested):
Let's get rid of all of the platformio stuff. Also, add all of the folders with stuff autogenerated by ceedling to a .gitignore.

We'll also look into a way to avoid having header files in the src folder.

matthewjegan (commented):
Applied changes to maintain project structure with separate header files.
Removed build files and autogenerated code.

a2k-hanlon (approved):

PR #17, opened 2021-09-11
GUI CAN Integration

a2k-hanlon:
Temporary PR to share some review


components/bms/gui_CAN_integration/Src/threading_class_revised.py
```diff
@@ -0,0 +1,139 @@
+from PyQt5.QtCore import QThread, pyqtSignal
```
a2k-hanlon, line 1: Hey Mischa, would you mind pointing out what has changed in this file? Since that branch hasn't been merged (we will need to do that), and your branch isn't based on the GUI branch, I don't have an indication in the diff of what actually changed.

Also, once we do merge the GUI branch to master, please make the change to the existing GUI source file instead of making a copy as you've done here.

tools/GUI_Nucleo_data_pipeline/Core/Src/main.c
```diff
@@ -0,0 +1,334 @@
...
+/* Private macro -------------------------------------------------------------*/
+/* USER CODE BEGIN PM */
+
+/* USER CODE END PM */
+
+/* Private variables ---------------------------------------------------------*/
+CAN_HandleTypeDef hcan;
+UART_HandleTypeDef huart2;
+
+CAN_FilterTypeDef sFilterConfig;
```
a2k-hanlon: `sFilterConfig` can be a local variable in `CAN_Config()`; as a rule of thumb, avoid global variables when possible! Just curious, what is the 's' at the beginning for? Also, be careful about adding code outside of the BEGIN and END comments :)

tools/GUI_Nucleo_data_pipeline/Core/Src/main.c
```diff
@@ -0,0 +1,334 @@
...
+  hcan.Init.Mode = CAN_MODE_NORMAL;
+  hcan.Init.SyncJumpWidth = CAN_SJW_1TQ;
+  hcan.Init.TimeSeg1 = CAN_BS1_1TQ;
+  hcan.Init.TimeSeg2 = CAN_BS2_1TQ;
+  hcan.Init.TimeTriggeredMode = DISABLE;
+  hcan.Init.AutoBusOff = DISABLE;
+  hcan.Init.AutoWakeUp = DISABLE;
+  hcan.Init.AutoRetransmission = DISABLE;
+  hcan.Init.ReceiveFifoLocked = DISABLE;
+  hcan.Init.TransmitFifoPriority = DISABLE;
```
a2k-hanlon, line 172: Check out the master board CAN settings (in CubeMX or in the `MX_CAN_Init()` function for that program) and make sure the settings here match up. All devices on the same CAN bus need to agree on the interface configuration. Note that the prescaler may be a different number because it depends on the MCU's clock frequency which might be different; it just needs to give the same data rate.

tools/GUI_Nucleo_data_pipeline/Core/Src/main.c
```diff
@@ -0,0 +1,334 @@
...
+  sFilterConfig.FilterBank = 1; //https://www.google.ca/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwjRrfzYluTxAhVXqJ4KHUWbB1wQFjAAegQIEBAD&url=https%3A%2F%2Fwww.st.com%2Fresource%2Fen%2Freference_manual%2Fdm00135183-stm32f446xx-advanced-armbased-32bit-mcus-stmicroelectronics.pdf&usg=AOvVaw09x0J9iGreyFuAMma4aTnE
+  //page 1058
+  sFilterConfig.FilterMode = CAN_FILTERMODE_IDMASK;
+  sFilterConfig.FilterScale = CAN_FILTERSCALE_32BIT;
+  sFilterConfig.FilterIdHigh = 0x623 << 5; //the filter and mask IDs will need to change
+  sFilterConfig.FilterIdLow = 0x0000;
+  sFilterConfig.FilterMaskIdHigh = 0x623 << 5;
+  sFilterConfig.FilterMaskIdLow = 0x0000;
+  sFilterConfig.FilterFIFOAssignment = CAN_RX_FIFO0;
+  sFilterConfig.FilterActivation = ENABLE;
```
a2k-hanlon: I can give you a hand with these once we have our protocol established. I looked at that link; good going finding a reference manual, but be careful because that's the manual for a set of STM32**F4** chips, whereas the team mostly uses STM32**F1** series. The BMS was going to use an F3 series chip (but we'll likely swap that for a F1 because we have a bunch)

components/bms/gui_CAN_integration/Inc/masterboard_data_collection.h
```diff
@@ -0,0 +1,46 @@
...
+BTM_PackData_t *pPackData;
+
+typedef struct
+{
+    CAN_TxHeaderTypeDef TxHeader;
+    uint8_t dataFrame[8];
+    uint32_t mailbox = 0; //check this
+} CAN_Message_t;
+
+#endif //this means this file will only be included once, so won't get error
```
a2k-hanlon: Always make sure your source files end with an empty line, it's just a convention.

components/bms/gui_CAN_integration/Src/masterboard_data_collection.c
```diff
@@ -0,0 +1,186 @@
+#include "masterboard_data_collection.h"
+
+CAN_Message_t *txInfo;
+uint8_t messageID_table[5] = {0x100, 0x200, 0x300, 0x400, 0x500};                                                                                           //placeholder values -- replace after talking to Matthew
+void (*pGetter_fns[5])(BTM_PackData_t *pPackData) = {Voltage_Retrieve, Temperature_Retrieve, SOC_Retrieve, Module_Status_Retrieve, System_Status_Retrieve}; //array of data getter functions
```
a2k-hanlon: This is clever but a little more complex than necessary here! I'd say a clearer implementation would just put these different function calls in a switch statement within `tx_data()`. In any case, this doesn't have to be a global array.

components/bms/gui_CAN_integration/Src/masterboard_data_collection.c
```diff
@@ -0,0 +1,186 @@
...
+
+                buffer_end += CAN_BYTES_SENT; //sent 8 bytes of data
+            }
+        }
+    }
+}
+
+void Voltage_Retrieve(BTM_PackData_t *pPackData)
+{
+    uint8_t collected_data[(RELEVANT_MODULES_PER_STACK * 2)]; //format: [MSBs, LSBs, MSBs, LSBs, ....]
```
a2k-hanlon: As far as I can tell you've populated this array correctly, so good job there! However, you'll lose it as soon as this function returns. What I suggest is to declare `collected_data[]` in `tx_data()` instead, and then you can pass it by reference to `Voltage_Retrieve()`. Same goes for the other retrieval functions.

components/bms/gui_CAN_integration/Src/masterboard_data_collection.c
```diff
@@ -0,0 +1,186 @@
...
+                k += 2;
+            }
+        }
+    }
+}
+
+void System_Status_Retrieve(BTM_PackData_t *pPackData)
+{
+    uint8_t collected_data[2];
+    uint16_t system_status16 = pPackData->PH_status;
```
a2k-hanlon: The PH in PH_status stands for placeholder... this isn't your problem but just as an FYI, this attribute of the packdata struct I believe is so far unused in the bms master board firmware so that'll have to be fixed.

components/bms/gui_CAN_integration/Inc/masterboard_data_collection.h
```diff
@@ -0,0 +1,46 @@
...
+#define CAN_BYTES_SENT 8
+#define TRANSMISSION_ATTEMPT_MAX 3
+#define BTM_NUM_DEVICES 2U
+#define BTM_NUM_MODULES 18
+#define CAN_MESSAGE_SERIES_LEN 5
+#define SEND_DATA 1 //true
+
+#define DATA_BYTES_MODULE 2
+
+CAN_HandleTypeDef *CAN_Handle; //then you could do &hcan = CAN_Handle;
```
a2k-hanlon: You'll have noticed that functions, typedefs and symbolic constants declared in the HAL CAN module have the 3 letter CAN_ prefix. I would suggest reserving this naming for things in the HAL only, not your own code, just to keep it clear what is defined in the HAL.

a2k-hanlon: This CAN handle is invalid unless it matches the CAN handle declared in main.c, so you need a way to communicate the main.c CAN handle to this code module. Having it passed in as a parameter from the caller of your functions, like the HAL functions do, is probably the safest option.

components/bms/gui_CAN_integration/Inc/masterboard_data_collection.h
```diff
@@ -0,0 +1,46 @@
...
+
+BTM_PackData_t *pPackData;
```
a2k-hanlon: This can be removed

components/bms/gui_CAN_integration/Src/masterboard_data_collection.c
```diff
@@ -0,0 +1,186 @@
...
+    while (SEND_DATA == 1)
+    {
+
+        for (int k = 0; k < CAN_MESSAGE_SERIES_LEN; k++)
+        {
+
+            buffer_pos = 0;
+            buffer_end = CAN_BYTES_SENT;
+            status = HAL_OK;
+            size_of_data = sizeof(collected_data) / sizeof(collected_data[0]); //bytes of data needing to be sent
```
a2k-hanlon: collected_data is not defined in this scope, see comment below

components/bms/gui_CAN_integration/Src/masterboard_data_collection.c
```diff
@@ -0,0 +1,186 @@
...
+    for (int i = 0; i < BTM_NUM_DEVICES; i++)
+    {
+        int k = 0;
+        for (int j = 0; j < RELEVANT_MODULES_PER_STACK; j++)
+        {
+            if (pPACKDATA->stack[i].module[j].enable == 1)
+            {
+                uint16_t module_voltage = pPackData->stack[i].module[j].voltage;
+                uint8_t module_voltage_MSBs = (uint8_t)(module_voltage >> 8); //shifting voltage data 8 bits to right, then casting uint16_t to uint8_t, discarding first 8 bits from left
+                uint8_t module_voltage_LSBs = (uint8_t)module_voltage;        //casting uint16_t to uint8_t, discarding first 8 bits from left, which were already saved in module_voltage_MSBs
```
a2k-hanlon: I found the comments here a little confusing in that they're both discarding the left (upper) bits, and I see what you mean (that's with the cast), but another way to look at it is one line is keeping the left (upper) half of the 16b voltage and the other line is keeping the right (lower) half of the 16b voltage.

tools/GUI_Nucleo_data_pipeline/Core/Inc/masterboard_data_collection.h
```diff
@@ -0,0 +1,43 @@
+#ifndef INC_MASTERBOARD_DATA_COLLECTION_H_
```
a2k-hanlon: Why is this file duplicated here?

components/bms/gui_CAN_integration/Src/masterboard_data_collection.c
```diff
@@ -0,0 +1,186 @@
...
+{
+    (txInfo + i)->TxHeader.StdId = messageID_table[i];
+    (txInfo + i)->TxHeader.ExtId = 0;
+    (txInfo + i)->TxHeader.IDE = CAN_ID_STD;
+    (txInfo + i)->TxHeader.RTR = CAN_RTR_DATA;
+    (txInfo + i)->TxHeader.DLC = 8;
+    (txInfo + i)->TxHeader.TransmitGlobalTime = 0;
+}
+
+//does this work, or would we need to init a entire new CAN_Messasge_t for each data type to send?
```
a2k-hanlon: I'd point out 2 things here: First, you need to put this code in a function. Second, there hasn't been a variable declared to store what txInfo points to. As for the comment, it depends on your implementation of tx_data().

tools/GUI_Nucleo_data_pipeline/Core/Inc/nucleo_recieve_data.h
```diff
@@ -0,0 +1,40 @@
...
+//characters GUI uses to classify type of incomming data
+
+//these CAN IDs are placeholder values until Matthew's project finalized
+ID_table_t VOLT_ID = {.CAN_ID = 0x100, .datatype = "v", .next = NULL};
+ID_table_t TEMP_ID = {.CAN_ID = 0x200, .datatype = "t", .next = NULL};
+ID_table_t SOC_ID = {.CAN_ID = 0x300, .datatype = "c", .next = NULL};
+ID_table_t MOD_STATUS_ID = {.CAN_ID = 0x400, .datatype = "s", .next = NULL};
+ID_table_t PACK_STATUS_ID = {.CAN_ID = 0x500, .datatype = "y", .next = NULL};
+
+CAN_RxHeaderTypeDef RxHeader;
```
a2k-hanlon: make this a local variable

tools/GUI_Nucleo_data_pipeline/Core/Inc/nucleo_recieve_data.h
```diff
@@ -0,0 +1,40 @@
...
+void hash_table_insert(ID_table_t *id_struct);
+
+//characters GUI uses to classify type of incomming data
+
+//these CAN IDs are placeholder values until Matthew's project finalized
+ID_table_t VOLT_ID = {.CAN_ID = 0x100, .datatype = "v", .next = NULL};
+ID_table_t TEMP_ID = {.CAN_ID = 0x200, .datatype = "t", .next = NULL};
+ID_table_t SOC_ID = {.CAN_ID = 0x300, .datatype = "c", .next = NULL};
+ID_table_t MOD_STATUS_ID = {.CAN_ID = 0x400, .datatype = "s", .next = NULL};
+ID_table_t PACK_STATUS_ID = {.CAN_ID = 0x500, .datatype = "y", .next = NULL};
```
a2k-hanlon: Move these to the .c file, and I advise against using all caps names for variables. It goes against C convention.

tools/GUI_Nucleo_data_pipeline/Core/Inc/nucleo_recieve_data.h
```diff
@@ -0,0 +1,40 @@
...
+
+CAN_RxHeaderTypeDef RxHeader;
+
+#define NEWLINE_CHAR "\n"
```
a2k-hanlon, line 16: use single quotes for a char, double quotes signify a string

tools/GUI_Nucleo_data_pipeline/Core/Src/nucleo_recieve_data.c
```diff
@@ -0,0 +1,48 @@
+#include "nucleo_recieve_data.h"
+
+CAN_HandleTypeDef *hcan;
+CAN_FilterTypeDef *sFilterConfig;
+uint8_t RxData[8];
```
a2k-hanlon: globals b gone

tools/GUI_Nucleo_data_pipeline/Core/Src/nucleo_recieve_data.c
```diff
@@ -0,0 +1,48 @@
...
+            startup = 0; //end startup sequence
+        }
+    }
+
+    if (startup == 0)
+    {
+        if (incomingID != currentID) //new datatype incoming
+        {
+            HAL_UART_Transmit(&huart2, (uint8_t *)&NEWLINE_CHAR, strlen(NEWLINE_CHAR), TIMEOUT); //newline char signals GUI that new data is incoming
+            HAL_UART_Transmit(&huart2, (uint8_t *)hash_table_lookup(incomingID), strlen(hash_table_lookup(incomingID)), TIMEOUT);
```
a2k-hanlon: Let's talk this over together the next time we meet. I need some clarification on the intended operation here.

a2k-hanlon (commented):
Thanks for your work here Mischa! I've shared some comments throughout this code. Let me know if you have questions and we can certainly talk it over further.

I adjusted the gitignore a little further - it had to go in the project folder you created rather than the root of the repo.

Also, please always try compiling your code to make sure that it at least builds without error, ideally before you push it, even if it's a work in progress. It makes reviewing it a little easier.

tools/GUI_Nucleo_data_pipeline/Core/Src/main.c
```diff
@@ -0,0 +1,328 @@
...
+
+  /* USER CODE BEGIN Init */
+
+  /* USER CODE END Init */
+
+  /* Configure the system clock */
+  SystemClock_Config();
+
+  /* USER CODE BEGIN SysInit */
+  CAN_config();
```
a2k-hanlon, line 88: This `CAN_config()` function should be called after `MX_CAN_Init()`

tools/GUI_Nucleo_data_pipeline/Core/Src/main.c
```diff
@@ -0,0 +1,328 @@
...
+  FilterConfig.FilterMode = CAN_FILTERMODE_IDMASK;
+  FilterConfig.FilterScale = CAN_FILTERSCALE_32BIT;
+  FilterConfig.FilterIdHigh = 0x623 << 5; //the filter and mask IDs will need to change
+  FilterConfig.FilterIdLow = 0x0000;
+  FilterConfig.FilterMaskIdHigh = 0x623 << 5;
+  FilterConfig.FilterMaskIdLow = 0x0000;
+  FilterConfig.FilterFIFOAssignment = CAN_RX_FIFO0;
+  FilterConfig.FilterActivation = ENABLE;
+
+  if (HAL_CAN_Init(&hcan) != HAL_OK)
```
a2k-hanlon, line 265: This is already in `MX_CAN_Config()` generated by CubeMX

tools/GUI_Nucleo_data_pipeline/Core/Src/main.c
```diff
@@ -0,0 +1,328 @@
...
+    Error_Handler();
+  }
+
+  //init CAN ID lookup table
+  init_hash_table();
+  hash_table_insert(&VOLT_ID);
+  hash_table_insert(&TEMP_ID);
+  hash_table_insert(&SOC_ID);
+  hash_table_insert(&MOD_STATUS_ID);
+  hash_table_insert(&PACK_STATUS_ID);
```
a2k-hanlon, line 292: It's probably best to call `HAL_CAN_Start()` as the final step of this function, once you have established all of the configurations (filters, interrupt, etc), rather than somewhere in the middle.

a2k-hanlon:
Closing this PR and deleting this branch as well as `user/matthewjegan/bms/gui_CAN_transceiver` since Mischa and I agree that it's not the direction we want to take for reading off of the car's CAN bus from a computer now that the team has a PCAN interface.

PR #18, opened 2021-10-02
migrated stm32f3 based project to stm32f103rc-based system

francisfu-embedded:


a2k-hanlon:
Hi Francis, thank you for your work here! I will look it over as soon as I can, but in the meantime, would you please restore the components/bms/bms_master/.gitignore file and remove the ignored files from the PR?

francisfu-embedded:
Hi Andrew. I just recovered the .gitignore file. Just wondering what do you mean by "remove the ignored files from the PR"? Do you mean to remove the object files in the debug folder?

a2k-hanlon:
Hi Francis, thank you for addressing that, but you have changed the wrong .gitignore file. We have more than 1 .gitignore file in this repository. You modified the .gitignore at the root of the repository rather than [components/bms/bms_master/.gitignore](https://github.com/UBC-Solar/firmware_v3/blob/master/components/bms/bms_master/.gitignore). Please revert the changes from your last commit and put back components/bms/bms_master/.gitignore.

Also, I see that what you put in that root .gitignore file is actually a bunch of HTML - I'm not sure what happened but please look over what you plan to commit before committing it.

To answer your question about removing the ignored files from the PR, I think this will make more sense once you see the bms_master folder .gitignore, but yes you are correct. The `Debug` folder within `components/bms/bms_master` needs to be removed. This folder was set to be ignored by git, but since a .gitignore file got deleted, you were able to inadvertently commit it with your migration changes.

francisfu-embedded:
THanks Andrew. Sorry that I put the wrong .gitignore file. This one should work

components/bms/bms_master/project.yml
```diff
@@ -1,108 +0,0 @@
----
```
a2k-hanlon: Please restore this file too

components/bms/bms_master/test/test_selftest.c
```diff
@@ -1,100 +0,0 @@
-#ifdef TEST
```
a2k-hanlon: Please restore this file too

a2k-hanlon:
No worries, thank you Francis! Thank you for adding back those 2 unit test files. I have now checked over your reconfiguration and it all looks good. Nice work! Thanks for your patience with my review.

There is just one more file I'd like you to make a change for - `components/bms/bms_master/.mxproject`. This one was around before your changes here but I noticed that it contains computer-specific absolute filepaths, and it's a CubeMX-generated file that we don't need in version control. So, please delete this file from the repo and add `.mxproject` to the bms_master `.gitignore`.

I am approving your PR now, so just feel free to merge it to master once you've made that change. Thanks again!

a2k-hanlon (dismissed):

francisfu-embedded:
oh it seems like I'm blocked because of the changes I've made. Can you approve again?

a2k-hanlon (approved):

francisfu-embedded:
merged

PR #21, opened 2022-01-24
Initial BMS State of Charge Estimation FW Implementation

forbes-choy:


components/bms/bms_master/Core/Inc/soc.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef __SOC_H
+#define __SOC_H
+
+#include "util.h"
+#include <stdint.h>
+#include "stm32f1xx_hal.h"
+#include "ltc6813_btm.h"
+
+/*============================================================================*/
+/* PUBLIC VARAIBLES */
```
a2k-hanlon: Remove unused heading

components/bms/bms_master/Core/Inc/soc.h
```diff
@@ -0,0 +1,55 @@
...
+/* PUBLIC CONSTANTS */
+
+//our cell's website: https://www.18650batterystore.com/products/sanyo-ncr18650ga
+//our cell's datasheet: https://cdn.shopify.com/s/files/1/0481/9678/0183/files/Sanyo-Spec-NCR18650GA.pdf?v=1605877395
+
+#define SOC_CELL_MIN_VOLTAGE 2.5 //voltage of a cell when fully discharged
+#define SOC_CELL_MAX_VOLTAGE 4.2 //voltage of a cell when fully charged
+#define SOC_CELL_DISCHARGE_EFFICIENCY 0.85 //according to http://large.stanford.edu/courses/2010/ph240/sun1/, but subject to (experimental) change
+#define SOC_CELL_CHARGE_EFFICIENCY 0.85 //according to http://large.stanford.edu/courses/2010/ph240/sun1/, but subject to (experimental) change
+#define SOC_MODULE_RATED_CAPACITY (3.45 * 13) // capacity [Ah] of each module; each cell has a typical capacity of 3450mAh and each module has 13 cells in parallel
```
a2k-hanlon: Nit: For easy reading, consider tab aligning defined values and then comments

components/bms/bms_master/Core/Inc/soc.h
```diff
@@ -0,0 +1,55 @@
...
+float SOC_moduleInit(float cell_voltage);
+void SOC_allModulesInit(BTM_PackData_t * pack);
+
+//functions require calling each time when CAN transmits a new current to update our SOC estimation
+float SOC_moduleEst(float last_SOC, uint32_t cell_voltage_100uV, int32_t current_reading, uint32_t total_time_elasped);
+void SOC_allModulesEst(BTM_PackData_t * pack, int32_t current_reading, uint32_t total_time_elasped);
+
+//private helper functions
+#ifdef TEST
+#include <stdio.h> //for debugging
```
a2k-hanlon: Put this include in the C file, wrapped in its own `#ifdef TEST` if necessary, so it's clear that this is a dependency of the code rather than it being buried down near the bottom of the header

components/bms/bms_master/Core/Inc/soc.h
```diff
@@ -0,0 +1,55 @@
...
+ *        coloumb counting
+ * 
+ * 
+ * @date 2021/11/09
+ * @author Edward Ma, Forbes Choy 
+ */
+
+
+#ifndef __SOC_H
+#define __SOC_H
```
a2k-hanlon: Just for consistency with other modules, please name the include guard
```suggestion
#ifndef INC_SOC_H_
#define INC_SOC_H_
```
Remember to update the comment by the `#endif` as well

components/bms/bms_master/platformio.ini
```diff
@@ -0,0 +1,23 @@
+; PlatformIO Project Configuration File
```
a2k-hanlon: I'm not sure if we should commit your platformio config for this project, since it is set up for CubeIDE. It doesn't seem to create a problem to have both, and having platformio set up does make working with this project in VS Code pretty nice. However, having 2 build methods is a bit redundant.

If we want to have platformIO set up, I think making that change in another PR is appropriate. So for now, please remove this file from this PR.

a2k-hanlon (approved):

PR #41, opened 2023-02-18
new function for DOC_COC

Tik-Hakobyan:


components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -7,18 +7,25 @@
...
 void CAN_send_current(int currentVal) {
+    int doc_coc = 0;
     uint32_t begin_tick = HAL_GetTick();
     uint32_t * pTxMailbox = 0;
     HAL_StatusTypeDef status;
     currentVal /= 100;
-    uint8_t data[] = {(uint8_t)currentVal, (uint8_t)(currentVal >> 8), 0, 0, 0, 0};
+
+    if (HAL_GPIO_ReadPin(OC_LATCH_SET_GPIO_Port, OC_LATCH_SET_Pin) == 1) doc_coc = 0;
+    else doc_coc = 1;
```
a2k-hanlon: This works, but just to share an alternative with you, it's also possible to do the following to accomplish the same thing a bit more concisely:

```suggestion
    doc_coc = !HAL_GPIO_ReadPin(OC_LATCH_SET_GPIO_Port, OC_LATCH_SET_Pin);
```

a2k-hanlon: I'd also prefer that you instead add a function parameter to this `CAN_send_current()` function for doc_coc and put this code calling `HAL_GPIO_ReadPin()` right before the call to `CAN_send_current()` in fsm.c.

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -7,18 +7,25 @@
...
+
+    uint8_t data[] = {(uint8_t)currentVal, (uint8_t)(currentVal >> 8), 0, (uint8_t) doc_coc, 0};
```
a2k-hanlon: First off, note that as of right now Mischa (@mjohal67) has changed the format of this message in the CAN table to this:

image

With Mischa's change as of now, DOC_COC is now bit 8, and the current reading has a range of -75A to 75A, placed in bits 0 through 7 (ie. the first byte).

Note that we always number bits starting from 0 ("zero-indexing").

What you have here places `doc_coc` in bit number 16, not 8 (or 9, as previously specified). Remember that each item in this `uint8_t` array is 8 bits. Don't make this change just yet though. I think DOC_COC is actually better off in bit 16 and we should change the format specification; this leaves room to place the LV current field in bits 8 through 15 (the entire 2nd byte).

A change I would like you to make here right now is to update the placement and scaling of the current value (`current_val`) according to Mischa's specification in the table. At the moment it is 16 bits (first 2 bytes), please change it to just one byte as in the table. Also please change the scale factor; note that the input to the function is a signed integer in 100ths of an Amp.

Finally, I'd like you to add a function description comment to this function. A function description comment looks like this:

```c
/**
 * @brief One-line description of what the function does
 *
 * More details if necessary...
 *
 * @param parameter_name Parameter description
 * @param another_parameter_name Another parameter description
 * @returns description of what the function returns, if there is a return value
 */
```

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -7,18 +7,25 @@
  */
 
 #include "can.h"
+#include "main.h"
 
 CAN_TxHeaderTypeDef current_message_header = {
-    0x624, 0, CAN_ID_STD, CAN_RTR_DATA, 6, DISABLE
+    0x450, 0, CAN_ID_STD, CAN_RTR_DATA, 3, DISABLE
```
a2k-hanlon: Thanks for changing the ID and length; note that the `data` array in `CAN_send_current()` should have the same size as specified here (3 bytes).

Also, please add `#define` statements after the `#includes` for this CAN ID (0x450) and message length (3).

components/ecu/ecu_firmware/Core/Src/can.c
mjohal67: After Andrew's comment on byte-alignment, I've adjusted the CAN protocol - you can find the latest version on the document, and in the photo below. 
image

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,31 @@
  * 
  * @date 2021/01/30
  * @author Blake Shular (blake-shular)
+
+ * @param currentVal the current flowing through the ECU
+ * @param doc_coc Discharge and Charge overcurrent status
```
a2k-hanlon: `@param` descriptions go in a comment specifically for the function, immediately above the function definition. Where you've put these is the file description, not a function description. Please take another look at my example in a previous comment. Just FYI, these function description comments are formatted according to a documentation tool called [Doxygen](https://www.doxygen.nl/manual/docblocks.html). We don't really use the tool though, we just use its formatting standard.

Also, please add more details to your descriptions. What are the units of `currentVal` supposed to be? What values are allowed for `doc_coc` and what does each mean?  The point of these descriptions is to inform someone else how to use this function correctly.

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,31 @@
...
  */
 
 #include "can.h"
+#include "main.h"
+
+#define CAN_ID 0x450
+#define MESSAGE_LENGTH 3
```
a2k-hanlon, line 13: Looking good 👍

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,31 @@
...
+
 
 CAN_TxHeaderTypeDef current_message_header = {
-    0x624, 0, CAN_ID_STD, CAN_RTR_DATA, 6, DISABLE
+    CAN_ID, 0, CAN_ID_STD, CAN_RTR_DATA, MESSAGE_LENGTH, DISABLE
 }; 
 
-void CAN_send_current(int currentVal) {
+void CAN_send_current(int currentVal, int doc_coc) {
```
a2k-hanlon, line 25: Your function description comment (with the `@param` lines) needs to go right above this line

components/ecu/ecu_firmware/Core/Src/fsm.c
```diff
@@ -438,7 +438,8 @@ void ECU_monitor () {
 
     // send CAN message with current values to BMS
     if (timer_check(MESSAGE_INTERVAL)) {   
-        CAN_send_current(ADC3_netCurrentOut(ADC3_getArrayCurrent(), ADC3_getMotorCurrent()));
+        int doc_coc = !HAL_GPIO_ReadPin(OC_LATCH_SET_GPIO_Port, OC_LATCH_SET_Pin);
+        CAN_send_current(ADC3_netCurrentOut(ADC3_getArrayCurrent(), ADC3_getMotorCurrent()), doc_coc);
```
a2k-hanlon, line 442: Also looking good

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,31 @@
...
     uint32_t begin_tick = HAL_GetTick();
     uint32_t * pTxMailbox = 0;
     HAL_StatusTypeDef status;
     currentVal /= 100;
-    uint8_t data[] = {(uint8_t)currentVal, (uint8_t)(currentVal >> 8), 0, 0, 0, 0};
+
+    uint8_t data[] = {(uint8_t)currentVal, 0, (uint8_t) doc_coc};
```
a2k-hanlon, line 31: This field placement matches the CAN table spec now, thanks for updating it

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,31 @@
...
-    0x624, 0, CAN_ID_STD, CAN_RTR_DATA, 6, DISABLE
+    CAN_ID, 0, CAN_ID_STD, CAN_RTR_DATA, MESSAGE_LENGTH, DISABLE
 }; 
 
-void CAN_send_current(int currentVal) {
+void CAN_send_current(int currentVal, int doc_coc) {
     uint32_t begin_tick = HAL_GetTick();
     uint32_t * pTxMailbox = 0;
     HAL_StatusTypeDef status;
     currentVal /= 100;
```
a2k-hanlon: You still need to change this scaling according to our discussion, copied below for reference:

> For the new scaling:
>
> - The current value field in the CAN message data is 8 bits wide
> - The range of currents that must be representable by that field is ±75A
> - Your input to the function is 100ths of an amp (ie. if currentVal is -125 then the current flowing is -1.25A)
>
> The maximum value for a signed 8 bit value is 0b01111111 = 0x7F = 127. You can therefore calculate the scale factor you need as (1A/100)*(127/75A)

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,38 @@
...
+#define MESSAGE_LENGTH 3
+
 
 CAN_TxHeaderTypeDef current_message_header = {
-    0x624, 0, CAN_ID_STD, CAN_RTR_DATA, 6, DISABLE
+    CAN_ID, 0, CAN_ID_STD, CAN_RTR_DATA, MESSAGE_LENGTH, DISABLE
 }; 
 
-void CAN_send_current(int currentVal) {
+/*
```
a2k-hanlon: Doxygen comments need 2 asterisks on the first line, see below. You should notice when you do this that VS Code will highlight certain parts of the comment differently.

Also, please add a one line description of what the function does at the top of this comment (also see below).

```suggestion
/**
 * @brief Transmit ECU current reading CAN message
 *
```

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,38 @@
...
+                        It takes the value of 0 if OC_LATCH_SET is 1, and 1 otherwise.
+ */
+
+void CAN_send_current(int currentVal, int doc_coc) {
     uint32_t begin_tick = HAL_GetTick();
     uint32_t * pTxMailbox = 0;
     HAL_StatusTypeDef status;
-    currentVal /= 100;
-    uint8_t data[] = {(uint8_t)currentVal, (uint8_t)(currentVal >> 8), 0, 0, 0, 0};
+    currentVal = currentVal * 127 / 7500;
```
a2k-hanlon: Looks good!

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,38 @@
  * 
  * @date 2021/01/30
  * @author Blake Shular (blake-shular)
- */
+
+*/
```
a2k-hanlon: Remove this extra line and realign the asterisks

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,38 @@
...
-void CAN_send_current(int currentVal) {
+/*
+ * @param currentVal    The current flowing through the ECU, in Amps.
+                        The range is ±75A. 
+                        Scaling factor for this range: (1A/100)*(127/75A) = 127/7500
+
+ * @param doc_coc       Discharge and Charge overcurrent status.
+                        It takes the value of 0 if OC_LATCH_SET is 1, and 1 otherwise.
+ */
+
```
a2k-hanlon: Remove the empty line between the comment and the function here

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -4,21 +4,38 @@
...
 CAN_TxHeaderTypeDef current_message_header = {
-    0x624, 0, CAN_ID_STD, CAN_RTR_DATA, 6, DISABLE
+    CAN_ID, 0, CAN_ID_STD, CAN_RTR_DATA, MESSAGE_LENGTH, DISABLE
 }; 
 
-void CAN_send_current(int currentVal) {
+/*
+ * @param currentVal    The current flowing through the ECU, in Amps.
+                        The range is ±75A. 
+                        Scaling factor for this range: (1A/100)*(127/75A) = 127/7500
```
a2k-hanlon: Couple things:
- It's not current flowing through the ECU, it's the net current charging or discharging the battery pack on the high voltage bus
- The input to the function is not in amps, it's in 100ths of an amp, hence the scale factor. You're supposed to describe the function input here, not the output.
- I'd suggest putting the line about the scaling factor in a regular comment within the function

a2k-hanlon (commented):
Almost there Tigran, thanks for your work here. Just a few more things to fix. Please mark the PR ready to merge (remove the draft status), and have @bhavitalodo take a look too for visibility once done!

a2k-hanlon (approved):
Good to go!

PR #49, opened 2023-05-18
BMS GUI CAN Transceiver

a2k-hanlon:
Creating a record of this branch before deletion.


a2k-hanlon:
Closing this PR and deleting this branch since Mischa and I agree that it's not the direction we want to take for reading off of the car's CAN bus from a computer now that the team has a PCAN interface.

PR #55, opened 2023-06-10
Telemetry - IMU RTOS Threads

ckpaz12:


components/tel/Core/Src/imu.h
MatthewChow03, line 1: Move to include folder. And not sure why Michael put this in `imu.h` instead of a c/h pair. If there's a reason then that's fine, but if not maybe move split things up to keep consistent style.

components/tel/Core/Src/main.c
```diff
@@ -0,0 +1,232 @@
...
+
+CAN_FilterTypeDef CAN_filter0;
+CAN_FilterTypeDef CAN_filter1;
+
+CAN_RxHeaderTypeDef can_rx_header;
+
+HAL_StatusTypeDef can_start;
+
+uint8_t current_can_data[8];
+HAL_StatusTypeDef rx_status;
```
MatthewChow03: This is pretty standard for our projects because we used superloops for the most part and then copied code over for subsequent projects. Since there is no longer CAN logic code in `main.c/h`, it would be nicer to encapsulate the CAN code all in `can.c` and `can.h` and then have a function call to setup anything necessary in main (which gets included from `can.h`). For example, after calling `MX_CAN_Init()` we call a single function `Can_Init()` which sets up the filters, starts CAN with HAL, and activates notifications for interrupts.

This just makes things easier to find and cleaner.

components/tel/Core/Src/main.c
```diff
@@ -0,0 +1,232 @@
...
+  MX_I2C1_Init();
+  /* USER CODE BEGIN 2 */
+
+  // <----- CAN set-up ------>
+  CanFilterSetup();
+  can_start = HAL_CAN_Start(&hcan);
+  assert_param(can_start == HAL_OK);
+
+  HAL_StatusTypeDef can_notification_status = HAL_CAN_ActivateNotification(&hcan, CAN_IT_RX_FIFO0_MSG_PENDING);
+  assert_param(can_notification_status == HAL_OK);
```
MatthewChow03: See previous comment on CAN

components/tel/Core/Src/imu.h
```diff
@@ -0,0 +1,95 @@
...
+#define DEVICE_ADDRESS ((0x6A)<<1)
+
+
+#define UART_TIMEOUT 1000
+
+void transmitData(uint8_t* data, uint16_t size)
+{
+	HAL_UART_Transmit(&huart3, data, size, UART_TIMEOUT);
+	HAL_Delay(1000);
+}
```
MatthewChow03: Whether you move these functions to be in a declaration + definition in a c/h file or not, make sure to add some function documentation. I don't remember what this is and it isn't called anywhere in the project. Remove?

components/tel/Core/Src/imu.h
```diff
@@ -0,0 +1,95 @@
...
+	// The value is expressed as a 16-bit word in two’s complement.
+	accel = (OUT_H_A << 8) | (OUT_L_A);
+
+	accel = (float) accel * 0.061; // datasheet pg10
+
+	return (float) accel;
+}
+
+/*
+	Set up for gyro/accel modes
```
MatthewChow03: Example documentation. This could be found by reading the datasheet for like half an hour and figuring things out, but it would be nice to have here in the code since it doesn't change. Like what is being initialized for the IMU etc.

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+    osDelay(KERNEL_LED_DELAY);
+  }
+}
+
+__NO_RETURN void readCANTask(void *argument) {
+  static HAL_StatusTypeDef rx_status;
+  static CAN_msg_t current_can_message;
+
+  while (1) {
+    // wait for thread flags to be set
```
MatthewChow03: Nit: More complete would be "wait for thread flags to be set in the CAN rx fifo0 interrupt callback" small things like these just save time for future you and other devs

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+
+      for (uint8_t i=0; i<8; i++) {
+        current_can_message.data[i] = current_can_data[i];
+      }
+
+      // priority=0, timeout=0
+      osMessageQueuePut(canMessageQueueHandle, &current_can_message, 0U, 0U);
+    }
+
+    HAL_CAN_ActivateNotification(&hcan, CAN_IT_RX_FIFO0_MSG_PENDING);
```
MatthewChow03, line 274: If you go to the HAL documentation you'll see a section with steps to use CAN rx with interrupts. It may be helpful to comment here that this allows the next interrupt to trigger again.

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+
+  while (1) {
+    // retrieve CAN message from queue
+    queue_status = osMessageQueueGet(canMessageQueueHandle, &can_message, NULL, osWaitForever);
+
+    if (queue_status != osOK){
+      osThreadYield();
+    }
+
+    uint8_t can_buffer[21] = {0};
```
MatthewChow03: Make some macros for these magic numbers. What does 21 represent? Then if it's not obvious why 21, next to the macro definition you could write something like how it's the max size and an example

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+      // can_stream[2+i] = 0xFFUL & (can_message.data[i]);
+        can_buffer[i + 10]= can_message.data[i];
+    }
+
+    // CAN DATA LENGTH: 1 ASCII character
+    uint8_t length = "0123456789ABCDEF"[ can_message.header.DLC & 0xFUL];
+    can_buffer[18] = length;
+
+    // NEW LINE: 1 ASCII character
+    can_buffer[19] = '\n';
```
MatthewChow03: Same as these, they could be macros or `CAN_BUFFER_LEN - 1`, or something else that makes sense

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+    addtoIMUQueue("G", "Z", gy_z);
+    addtoIMUQueue("A", "X", ax_x);
+    addtoIMUQueue("A", "Y", ax_y);
+    addtoIMUQueue("A", "Z", ax_z);
+
+    osDelay(1000);
+  }
+
+  // In case we accidentally exit from task loop
+  osThreadTerminate(NULL);
```
MatthewChow03: How is this possible?

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+
+  //Convert to ASCII
+  uint8_t c_H = "0123456789ABCDEF"[(c >> 4) & 0xFUL];
+  uint8_t c_L = "0123456789ABCDEF"[ c & 0xFUL];
+
+  uart_tx_status = HAL_UART_Transmit(&huart3, &c_H, 1, 1000);
+  uart_tx_status = HAL_UART_Transmit(&huart3, &c_L, 1, 1000);
+}
+
+__NO_RETURN void StartReadIMU(void *argument){
```
MatthewChow03: As with function documentation, same for all tasks, short descriptions would be helpful.

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+    if (osMessageQueueGetCount(imuMessageQueueHandle) == 0) {
+        osThreadYield();
+    }
+
+    imu_queue_status = osMessageQueueGet(imuMessageQueueHandle, &imu_message, NULL, osWaitForever);
+
+    if (imu_queue_status != osOK){
+      osThreadYield();
+    }
+    uint8_t imu_buffer[9] = {0};
```
MatthewChow03: Same as CAN buffer, 9 would be better as a macro. It's a magic number without an initially apparent reason as to why it is size 9.

components/tel/Core/Src/freertos.c
```diff
@@ -0,0 +1,462 @@
...
+
+  /* USER CODE BEGIN RTOS_THREADS */
+
+  // kernelLEDTaskHandle = osThreadNew(kernelLEDTask, NULL, &kernelLEDTask_attributes);
+  
+  readCANTaskHandle = osThreadNew(readCANTask, NULL, &readCANTask_attributes);
+  transmitMessageTaskHandle = osThreadNew(transmitMessageTask, NULL, &transmitMessageTask_attributes);
+
+  readimuHandle = osThreadNew(StartReadIMU, NULL, &readimu_attributes);
+  transmitIMUHandle = osThreadNew(StartTransmitIMU, NULL, &transmitIMU_attributes);
```
MatthewChow03: nameTask could be more clear than StartName. This also helps quickly differentiate between tasks and helper functions which both exist in this file.

MatthewChow03 (approved):
Great work! Can you add a comment summarizing the changes/additions and the current state of the TEL at this PR. I think it could be helpful to look at in the future to know what works and what has been tested at this point and with how much certainty.

Some comments for code styling and clarity but nothing major. Take what you want and resolve the comments, leave other things you think are good as is. 

Preemptive approval. Make sure to look over all the comments before resolving them individually and then good to merge.

PR #59, opened 2023-06-14
Kyle's MCB Code

MatthewChow03:
All of the newest MCB code from Kyle to add pots, updated state machine, and refactored RTOS architecture. 

_Note: I have modified master's `mcb/mcb_main` project to change MCU from STM32F103RCT6 to STM32F103RBT6. All of Kyle's code has been copied into `mcb/mcb_main` with version history kept. The diff between Mihir's MCB code and Kyle's code all comes up in one commit, but the version history persists - unlike prior attempts or adding the project to a new folder which loses tracking and shows no comparison to past code._

Steps to reproduce MCU .ioc change and code replacement (add to docs):
- Remove all files from the `components/project_name` folder (new projects require empty folders)
- *Make a new project in the empty folder (this will require you to generate code)
  - If from scratch, configure all the hardware on the MCU to match the specs (use the same naming for pinout to allow source code to transfer) 
  - If from existing .ioc, everything should be done and good to go
- *Copy back all .c and .h files from the original project to the `Core` folder
- Regenerate code from .ioc
- Build code again
- Make sure the project specific .gitignore file is in the new `components/project_name` folder because it got deleted in the first step

_Some of these steps can be combined, but I would suggest committing after each * step above separately to save the history and easily see what was modified_


MatthewChow03:
Note CI broken since modified the MCB directory. May look into fixing that later.

MatthewChow03:
TODO
- Check this code works on target same as Kyle's most recent
- Redo this process to show Kyle and clean up steps
- Make a new PR and merge to master
- Delete all other MCB branches and close other MCB PRs
- Document this process of swapping MCUs while keeping version history

PR #60, opened 2023-06-15
AMB voltage and current models with CAN TX

ishanjoshi23:
Current AMB FW Updates: Voltage and Current models implemented; still waiting on Thermistors to implement temperature models. 
CAN support is also added.


components/mcb/mcb_main/.project
MatthewChow03, line 1: Not sure why this is here. Probably should remove.

.project
MatthewChow03, line 1: Remove

MatthewChow03 (dismissed):
Lgtm, pre-emptive approval. You might want to add a `.gitignore` inside `components/amb`.

.project
ishanjoshi23: Sure, will remove it

components/mcb/mcb_main/.project
ishanjoshi23: Removed it

a2k-hanlon:
The AMB project is a CubeIDE project wit a `.project` file of its own, so I don't think ignoring all .project files is a good idea. Removing the stray mcb .project file should be suffice.

MatthewChow03:
@ishanjoshi23 I meant something like the project specific .gitignore like https://github.com/UBC-Solar/firmware_v3/blob/master/components/mcb/mcb_main/.gitignore

But I didn't realize you have things set up. For example, you have a git ignore inside the Debug folder. If the amb setup ignores all the random build files then it should be fine

a2k-hanlon:
@MatthewChow03 Ishan did make a project specific gitignore `components/amb/.gitignore`. But, it's contents are just `*.project`. I wouldn't recommend that because it ignores the amb project's `.project` file.

If you want to go for consistency with other projects, keep `components/amb/.gitignore`, remove `components/amb/Debug/.gitignore`, and put something like this in `components/amb/.gitignore`:

```
# Build outputs
Debug/
Release/

# Project log
.mxproject

# Debug configuration
*.cfg
*.launch

```

ishanjoshi23:
> @MatthewChow03 Ishan did make a project specific gitignore `components/amb/.gitignore`. But, it's contents are just `*.project`. I wouldn't recommend that because it ignores the amb project's `.project` file.
> 
> If you want to go for consistency with other projects, keep `components/amb/.gitignore`, remove `components/amb/Debug/.gitignore`, and put something like this in `components/amb/.gitignore`:
> 
> ```
> # Build outputs
> Debug/
> Release/
> 
> # Project log
> .mxproject
> 
> # Debug configuration
> *.cfg
> *.launch
> ```

ishanjoshi23:
@MatthewChow03 @a2k-hanlon I've removed the .gitignore from /Debug/ and included the above mentioned contents in the gitignore in /components/amb/.gitignore

a2k-hanlon:
Looks good! You may want to also remove, and commit the removal of, the files that are now ignored including the `.launch` file and the contents of the `Debug` folder

ishanjoshi23:
Ok, I have removed those.

MatthewChow03 (approved):

PR #73, opened 2023-12-06
User/aarjav jain101/tools/sd logger

AarjavJain101:
Created the'sdCardLib' library in the libraries folder of firmware_v3 and created an example code snippet in the SD_Logger tools project to use the abstracted function sd_open, sd_append, and sd_close.


ishanjoshi23 (commented):
There is a bunch of files here that are not supposed to be merged in (e.g., bunch of files that are compiled for the MCB, something to do with the AMB, and a bunch of  .cyclo files (should not be committed). Can you please clean up the branch so it only has those that we intend to add in?

PR #79, opened 2024-01-20
User/kcgroulx/mcb/dev

kcgroulx:
Tested MCB code


ishanjoshi23:
It would be great if we could add a README.md for this specific component which summarizes everything from the DR + codebase

ishanjoshi23:
Can we try to rename the project from `mcb_main` to `MCB_v1_0`?

ishanjoshi23 (approved):
Looks good. Will not delete the source branch.

PR #83, opened 2024-02-24
User/ishanjoshi23/did/did v3

kcgroulx:


ishanjoshi23:
Can we try to rename the project from `DashboardCompetitionRemake` to `DID_V3` . The former name is inherited from Daybreak and wasn't ever changed. Let me know if you have trouble with this.

ishanjoshi23:
Also, it would be great if we could add a README.md for this specific component which summarizes everything from the DR + codebase

ishanjoshi23 (approved):
Looks good. Firmware has been tested on DID board for vaildation.

PR #86, opened 2024-03-23
Added RTC Timestamp CAN message

kcgroulx:
Added 0x751 RTC timestamp CAN message. Gets the time as a Unix epoch time stamp and converts it to an array of 8 bytes.
Tested and verified with PCAN.

  
image


ishanjoshi23 (approved):
Looks good!

PR #88, opened 2024-03-30
feature:MDI_v1_0

cenamora:
PR for the MDI firmware which handles message and signal interfacing between the MCB and the Mitsuba Motor Controller.  

New PR since the previous PR was deleting main branch components. Incorporated the feedback from Ishan on the previous PR attempt.


ishanjoshi23 (approved):
Looks good to me

PR #89, opened 2024-04-03
MCB diagnostics

kcgroulx:
Added mcb diagnostics CAN message sent every 100ms.
Bits 0-15: pedal ADC reading
Bit 16: pedal ADC out of range
Bit 17: regen on
Bit 18: cruise on


ishanjoshi23 (approved):
Looks good to me

PR #96, opened 2024-05-08
feat: TEL firmware with support for CAN, IMU, and GPS messages

ishanjoshi23:


ishanjoshi23:
feat: TEL firmware with support for CAN, IMU, and GPS messages for release v0.3.

Key features:

- Rx CAN data using interrupts
- Poll IMU using I2C for GYRO and ACCEL data.
- Poll GPS for positioning and time data.
- Sync RTC with GPS time if internal RTC clock has reset (to preset Jan 1, 2000)
- Utilizes RTOS (CMSIS v1) for scheduling tasks
- Radio Tx using USART 1 at baudrate 230,400 bits/second.

> Note: the SD logger FATFS middleware and SPI peripheral initialization exists, but all FATFS & SD library calls are commented out for this release as SD Logger is not working reliably.

PR #102, opened 2024-05-20
del: Remove original (wrong) TEL project

ishanjoshi23:
This TEL project was originally build for the wrong MCU. Removing it now that we have a stable version of TEL on master.


mjohal67 (approved):
Approved.

PR #105, opened 2024-05-22
Project Renames

kcgroulx:
Project renames / cleanup

components/ambmarch6      -> components/amb
components/did_v3               -> components/did
components/mcb/mcb_v1_0 -> components/mcb
components/tel_v2                -> components/tel

removed unnecessary files from git tracking.


ishanjoshi23 (changes requested):
There is a `tel` and `tel_v2` folder now. Master only has `tel_v2` which should be called `tel`. Please also confirm that all projects can successfully compile after the name change.

PR #109, opened 2024-05-24
Sending RTC Timestamp with Memorator

AarjavJain101:
**Purpose**:
﻿* The 't' programming language script for sending the RTC timestamp has been created and tested to be correct based on print statements within the code itself. This script will provide us with the most accurate timestamps for all the other data being received by the TEL board. ﻿Now it needs to be tested by running it with the system.﻿
image

**How it Works**: [ASSOCIATED MONDAY POST](https://ubcsolar.monday.com/boards/3313681052/pulses/6693946997?term=memor&termColumns=XQAAAAJvAAAAAAAAAABBKoJ4MRmOSvlPTDV2Qow42wcHf67VzsBdKlCbBGsDUWUe9ZBNkdE8anb-D9z8-b1wBiZRkBInPvJQFM9-LnxSVAbsA45GP2YL3_0uWUA)
* Currently the script has 1 timer which goes off every 500ms (will change for practical use). 
* When this minute is up, the RTC message is sent with **ID = 0x751** **(PICTURE IS WRONG)**, data length = 8, and data[0] = seconds, data[1] = minutes, data[2] = hours, data[3] = day of month, data[4] = month, data[5] = year since 2000 (because thats how STMs HAL_RTC_DATE object interprets the year), data[6] = daylight savings flag.
* To get the timestamp the function ﻿timeGetDate ﻿is used which uses the ﻿tm struct:
image
* ~freeRTOS.c changed to check if the received message ix 0x751. If it is then it will extract the data from the message and set the time and date. Additionally, I modify the data to be an epoch timestamp in seconds as double because otherwise I will have to create 7 signals to make the value more interpretable (up for discussion, see Monday post on implementation details).~ ****
* ~Commented out~ **DELETED** transmit_RTC_timestamp function registration in freeRTOS.c and Sync_RTC_with_GPS function in main.c. 
* Removed gps_syn_fail and gps_fail fields from the TEL diagnostics and replaced them with gps_fix check instead. 
* Added RTC_reset check in freeRTOS read_CAN_task.
* Memorator has 0 delay before sending out periodic messages. It already has a 100ms delay in practice when the battery starts up and gives power to TEL first then memorator 100ms after.

﻿﻿**Notes:**
* I used the "﻿kvaser_t_programming_language" pdf documentation from their website to create this script (93 pages). 
* **Please see the Monday update for details**.

﻿
﻿﻿**TODO Before Merge:**
* ~**STEP 1**: Need to use this script on the memorator and ﻿confirm ﻿the 0x750 messages are being at least received by the TEL board.~
* ~**STEP 2:** Once we validate that the message is being received, we can move on to updating/syncing the TEL board time using HAL_setTime and HAL_setDate functions.~
* ~*STEP 3:** ﻿﻿Finally, we will validate that sunlink is receiving the 0x751 message correctly and the other messages have correct time stamps.~
* ~**STEP 4:** Need to run script in logging mode and see if powering the memorator will automatically run the script. See Monday post for details on how to get this to work with a configuration file and a python script to download the `sendRTC.txe` executable and XML config file onto the memorator.~
* **STEP 5:** Perform testing with the memorator by at least running the system for 20 minutes and then doing a log upload on sunlink. In the process, the values of `rtc_reset` and `gps_fix` will need to be checked so testing this outside to get a fix and inside to ensure the fail case works as well is important.


ishanjoshi23:
New branch created with fixes. Closing this PR

PR #110, opened 2024-05-25
Feat: TEL IWDG

kcgroulx:
Added independent watchdog to TEL project.
- IWDG is set to reset at approx 500ms.
- Refreshed every 100ms in the default task and every time a CAN message is received.


kcgroulx:
Tested.

PR #112, opened 2024-05-28
Fix: Non-blocking IMU initialization 

kcgroulx:


ishanjoshi23 (commented):
This was intended to fix CAN messages not being sent from the TEL board, but this was due to another issue, so closing this PR.

PR #113, opened 2024-05-31
Fix: Disable iwdg in DEBUG build

kcgroulx:
iwdg resets the board when we use the debugger to pause or stop at breakpoints.
Added a preprocessor that disables the iwdg when using the debugger.


ishanjoshi23 (approved):
Looks good. This was tested May 30, 2024.

PR #120, opened 2024-06-07
Added debugger cpu load

kcgroulx:
Added an optional cpu load debugger message. 
Must build with debug and set TEL_PRINT_CPU_LOAD to 1 in main.h


ishanjoshi23:
Closing PR as we don't want to keep this in production firmware.

PR #122, opened 2024-06-09
Refactor: Cleaning and Clarifying freertos.c code

AarjavJain101:
[**Accompanying Monday Update**](https://ubcsolar.monday.com/boards/3313681052/pulses/6789943092?term=refac&termColumns=XQAAAAJqAAAAAAAAAABBKoJ4MRmOSvlPTDV2OJa42wcHf67VzsBdKlCbBGsDUWUe9ZBNkdE8anb-D9z8-b1wBiZRkBInP8qpXmFBPkp7aiHcQVeT__9qyoAA)

**Purpose**:
* Our current repository code for TEL is messy in the sense that many constants/parameters are passed without being defined with symbolic constants and as such it is difficult to understand why certain values are chosen and also makes it difficult to make bulk changes to functionality that uses those parameters.
* Also, by cleaning up the code, using better naming conventions to establish a standard, and also to create functions to abstract logic and promote code resuse we set ourselves up for a strong foundation when adding new features like API mode.

**Status**:
* Currently the logic is in place for the refactor of freertos.c, however, there are still bugs that need to be fixed and tests that need to be conducted to ensure the refactor did not break any of the production code. 
* Please see the accompanying monday update at the top for details on debugging issues and the current problem.

### DIFF
--- a/components/tel/Core/Inc/rtc.h
+++ b/components/tel/Core/Inc/rtc.h
@@ -37,6 +37,14 @@ extern "C" {
 extern RTC_HandleTypeDef hrtc;
 
 /* USER CODE BEGIN Private defines */
+
+#define TIMETYPEDEF_SECONDS_IDX     0
+#define TIMETYPEDEF_MINUTES_IDX     1
+#define TIMETYPEDEF_HOURS_IDX       2
+#define TIMETYPEDEF_DAY_IDX         3
+#define TIMETYPEDEF_MONTH_IDX       4   
+#define TIMETYPEDEF_YEAR_IDX        5
+
 /* USER CODE END Private defines */
 
 void MX_RTC_Init(void);
--- a/components/tel/Core/Inc/rtc.h
+++ b/components/tel/Core/Inc/rtc.h
@@ -47,7 +47,7 @@ double get_current_timestamp();
 double convertToEpochTime(RTC_TimeTypeDef *sTime, RTC_DateTypeDef *sDate, double milliseconds);
 int lastDayOfMonth(int month, int year);
 bool checkAndSetRTCReset();
-void sync_memorator_rtc(CAN_msg_t* rx_CAN_msg);
+void RTC_check_and_sync_rtc(CAN_msg_t* rx_CAN_msg);
 
 /* USER CODE END Prototypes */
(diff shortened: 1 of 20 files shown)


components/tel/Core/Src/can.c
```diff
@@ -1,272 +1,355 @@
...
+  CAN_RxHeaderTypeDef can_rx_header;
+  uint8_t can_data[8];
+
+
+  /* Get CAN message */
+//  while(HAL_CAN_GetRxFifoFillLevel(hcan, CAN_RX_FIFO0) != 0) {
+    HAL_CAN_GetRxMessage(hcan, CAN_RX_FIFO0, &can_rx_header, can_data);  // TODO: Put can_rx_header and can_data into a data structure able to be accessed in the freertos task
+  //  printf("%d\n\r", HAL_CAN_GetRxFifoFillLevel(hcan, CAN_RX_FIFO0));
+    /* Put CAN message in the Queue */
+    CAN_msg_t *new_CAN_msg;
```
ubcsolar-admin: Variable should be declared at top of function.

components/tel/Core/Src/can.c
```diff
@@ -1,272 +1,355 @@
...
+    new_CAN_msg = osPoolAlloc(CAN_MSG_memory_pool);
+    new_CAN_msg->header = can_rx_header;
+    for(int i = 0; i < 8; i++) {
```
ubcsolar-admin: "8" should be \#defined.

components/tel/Core/Inc/rtc.h
```diff
@@ -1,59 +1,67 @@
...
+#define TIMETYPEDEF_HOURS_IDX       2
+#define TIMETYPEDEF_DAY_IDX         3
+#define TIMETYPEDEF_MONTH_IDX       4   
+#define TIMETYPEDEF_YEAR_IDX        5
+
+/* USER CODE END Private defines */
+
+void MX_RTC_Init(void);
+
+/* USER CODE BEGIN Prototypes */
```
ubcsolar-admin, line 52: All public functions in this file should be prefixed with RTC_xxx

PR #123, opened 2024-06-09
Feat: runtime counter for TEL diagnostic message

AarjavJain101:
[**Accompanying Monday Update**](https://ubcsolar.monday.com/boards/3313681052/pulses/6798409256)

See the monday update for the purpose of this feature. Essentially, it provides more debugging information that is sent every second. The point of having something updated independent of the peripherals on the board and the watchdog is that it tells you consecutive runtime and thus will explicitly tell you whether a power cycle occurred. **Importantly, if the power cycle occurs briefly and the other diagnostic information did not detect it in that time, the runtime counter will be guaranteed to reset no matter how quick the power cycle was; as long as the board reset in any way so that `main.c` is run again**.

**Changes**:
* Added a 1 second TIM2 interrupt based timer in which I ~increment the tel_diagnostic.runtime_counter uint16_t variable~ **call the `HEARTBEAT_handler()` function to send the diagnostic message.** 
* Added a SEND_HEARTBEAT.md doc that explains how to use the `heartbeat.h` file so that other boards can use this library. 

**Work Before PR**:
* Need freertos refactor ~changes~ **merge** to then ~edit the transmit diagnostic msg task so that the runtime counter is part of the message~ **use the `can.c` function to create the diagnostic data array in `main.c` and pass it to the `HEARTBEAT_handler()` function.**
* Need to update DBC and CAN ID table for this message.
* Need to delete the `transmit_diagnostic_task` because the heartbeat handler will take care of diagnostics from now on.


ishanjoshi23:
Not merging this as this is outdated, but some code will be pulled from this branch in generalizing an API.

PR #124, opened 2024-06-12
added heartbeat message

kcgroulx:
MR adds a heartbeat message for the DID with CANID 0x580 and a cadence of 1000ms.
Message is 2 bytes long and consists of a uint16_t counter that is incremented with every message.


ishanjoshi23:
This will change once we implement the standard diagnostic API

PR #126, opened 2024-06-15
Refactor: Testing, Cleaning and Clarifying freertos.c code

AarjavJain101:
[**Accompanying Monday Update**](https://ubcsolar.monday.com/boards/3313681052/pulses/6789943092?term=refac&termColumns=XQAAAAJqAAAAAAAAAABBKoJ4MRmOSvlPTDV2OJa42wcHf67VzsBdKlCbBGsDUWUe9ZBNkdE8anb-D9z8-b1wBiZRkBInP8qpXmFBPkp7aiHcQVeT__9qyoAA)

**Purpose**:
* Our current repository code for TEL is messy in the sense that many constants/parameters are passed without being defined with symbolic constants and as such it is difficult to understand why certain values are chosen and also makes it difficult to make bulk changes to functionality that uses those parameters.
* Also, by cleaning up the code, using better naming conventions to establish a standard, and also to create functions to abstract logic and promote code resuse we set ourselves up for a strong foundation when adding new features like API mode.

**Status**:
* ~Currently the logic is in place for the refactor of freertos.c, however, there are still bugs that need to be fixed and tests that need to be conducted to ensure the refactor did not break any of the production code.~
* ~Please see the accompanying monday update at the top for details on debugging issues and the current problem.~
* **INTEGRATION TESTING DONE. READY FOR REVIEW AND MERGE**

### DIFF
--- a/components/tel/Core/Inc/utils.h
+++ b/components/tel/Core/Inc/utils.h
@@ -1,0 +1,25 @@
+#include <stdint.h>
+
+/**
+ * @brief Union to convert float to bytes
+*/
+typedef union Utils_FloatBytes_t {
+    float float_value;
+    uint8_t bytes[4];
+} Utils_FloatBytes_t;
+
+
+/**
+ * @brief Union to convert double to bytes
+*/
+typedef union Utils_DoubleBytes_t {
+	double double_value;			/**< Double value member of the union. */
+	uint64_t double_as_int;			/**< 64 bit in member of union. */
+} Utils_DoubleBytes_t;
+
+void UTILS_zeroArray(uint8_t* array, uint8_t size);
+
+/**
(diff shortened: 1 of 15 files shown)


components/tel/Core/Inc/utils.h
```diff
@@ -0,0 +1,25 @@
+#include <stdint.h>
```
kcgroulx, line 1: Instead of copy-pasting this file across multiple projects, we can move it to ./libraries/ 
Multiple projects can access it using relative file path. IE #include "../../../../libraries/utils/utils.h"

kcgroulx: In the future, we can include the libraries directory so we dont need to use a long file path.

ishanjoshi23:
Merged original PR

PR #131, opened 2024-06-17
Packing messages as XBee API Frames

EvanO12:
As part of implementing api_packaging, I made the following changes:

1. Added RADIO_tx_API_Accumulator function. This function takes radio_buffers (can messages) from RADIO_tx_CAN_msg and stores them in an api_buffer.

2. Added RADIO_tx_API_Packager function. Once a set amount of can messages are in an api_buffer, this function is called, packages all the can messages in the api_buffer, and transmits the api_package over uart.

3. Added Xbee.c and .h files. Xbee.c contains functions for packaging xbee api packages, (and in the future, handling any return statuses, and perhaps AT commands). Xbee.c contains solely functions proprietary to the xbee api packaging process - that is, if we ever switch radio module brands, this file could be removed and replaced with a different file, and the rest of radio.c would be largely untouched.

4. Some things xbee.c does right now is calculate checksum, and length fields, as well as set up the static api overhead (destination addresses, etc).
I have not yet tested my changes, but I will do so soon in a standalone file. Let me know what you guys think.

### DIFF
--- a/components/tel/Core/Inc/radio.h
+++ b/components/tel/Core/Inc/radio.h
@@ -14,6 +14,7 @@
 #include "can.h"
 #include "usart.h"
 #include "rtc.h"
+#include "xbee.h"
 
 /* RADIO BUFFER BYTE LENGTHS */
 #define CAN_BUFFER_LEN                      24
@@ -42,6 +43,17 @@
 
 /* TIMING CONSTANTS */
 #define CAN_TRANSMIT_TIMEOUT                1000                // 1 second timeout
+
+/*XBee API Packet CONSTANTS*/
+#define API_BUFFER_SIZE 300 //number of bytes in api_buffeer array
+#define API_PACKET_SIZE 10 //number of messages in each api packet
+#define API_OVERHEAD_SIZE 18 //number of bits in api Overhead
+
+
+
+
+
+
 
 void RADIO_tx_CAN_msg(CAN_Radio_msg_t *tx_CAN_msg);
 
--- a/components/tel/Core/Inc/radio.h
+++ b/components/tel/Core/Inc/radio.h
@@ -16,6 +16,7 @@
 #include "rtc.h"
 #include "xbee.h"
 
+
 /* RADIO BUFFER BYTE LENGTHS */
 #define CAN_BUFFER_LEN                      24
 #define GPS_MESSAGE_LEN                     200
@@ -45,9 +46,9 @@
 #define CAN_TRANSMIT_TIMEOUT                1000                // 1 second timeout
 
 /*XBee API Packet CONSTANTS*/
-#define API_BUFFER_SIZE 300 //number of bytes in api_buffeer array
-#define API_PACKET_SIZE 10 //number of messages in each api packet
-#define API_OVERHEAD_SIZE 18 //number of bits in api Overhead
+#define API_BUFFER_SIZE 					300 			    //number of bytes in api_buffer array
+#define API_MAX_MSGS  						10 					//number of messages in each api packet
+#define API_OVERHEAD_SIZE 					20 					//number of bytes in api Overhead
 
 
 
--- a/components/tel/Core/Inc/xbee.h
+++ b/components/tel/Core/Inc/xbee.h
@@ -11,35 +11,41 @@
 #define INC_XBEE_H_
 
 #include <stdint.h>
+#include <radio.h>
 
 /*XBee API Packet Overhead*/
-#define START_DELIMITER 0x7E
-#define FRAME_TYPE 0x10
-#define FRAME_ID 0x01
-#define BIT_ADDRESS_16_LOW 0xFE
-#define BIT_ADDRESS_16_HIGH 0xFE
-#define BROADCAST_RADIUS 0x00
-#define OPTIONS 0x00
-#define TYPE 0x00
-#define BIT_ADDRESS_16_LOW 0xFE
-#define BIT_ADDRESS_16_HIGH 0xFE
+#define START_DELIMITER 					0x7E
+#define FRAME_TYPE 							0x10 				//0x10 is a transmit request
+#define FRAME_ID 							0x01
+#define BIT_ADDRESS_16_LOW 					0xFE
+#define BIT_ADDRESS_16_HIGH 				0xFE
+#define BROADCAST_RADIUS 					0x00
+#define OPTIONS 							0x00
+#define TYPE 								0x00
+#define BIT_ADDRESS_16_LOW 					0xFE
+#define BIT_ADDRESS_16_HIGH 				0xFE
+#define CAN_TYPE 							0x00
 
-#define CHECKSUM_LENGTH_SIZE 4 //size of fields in api packet that are excluded in checksum and length field calculations
+/*XBEE API PACKET Calculation Constants*/
+#define MASK_8_BITS 						0xFF
+#define UINT8_MAX_SIZE 						0xFF
+#define BIT_ADDRESS_64_LENGTH 				8
+#define CHECKSUM_LENGTH_SIZE 				4					//size of fields in api packet that are excluded in checksum and length field calculations
 
 /*XBee API Packet Positions*/
-#define START_DELIMITER_POSITION 0
-#define MSB_POSITION 1
-#define LSB_POSITION 2
-#define FRAME_TYPE_POSITION 3
-#define FRAME_ID_POSITION 4
-#define BIT_ADDRESS_64_START_POSITION 5
-#define BIT_ADDRESS_16_LOW_POSITION 13
-#define BIT_ADDRESS_16_HIGH_POSITION 14
-#define BROADCAST_RADIUS_POSITION 15
-#define OPTIONS_POSITION 16
-#define TYPE_POSITION 17
-#define LENGTH_POSITION 18
-#define MESSAGE_DATA_START_POSITION 19
+#define START_DELIMITER_POSITION 			0
+#define MSB_POSITION 						1
+#define LSB_POSITION 						2
+#define FRAME_TYPE_POSITION 				3
+#define FRAME_ID_POSITION 					4
+#define BIT_ADDRESS_64_START_POSITION 		5
+#define BIT_ADDRESS_16_LOW_POSITION 		13
+#define BIT_ADDRESS_16_HIGH_POSITION 		14
+#define BROADCAST_RADIUS_POSITION 			15
+#define OPTIONS_POSITION 					16
+#define TYPE_POSITION 						17
+#define LENGTH_POSITION 					18
+#define MESSAGE_DATA_START_POSITION 		19
 
 
 
--- a/components/tel/Core/Inc/xbee.h
+++ b/components/tel/Core/Inc/xbee.h
@@ -1,0 +1,50 @@
+
+/**
+ *  @file xbee.h
+ *  @brief header file for xbee.c. Define xBee API packet positions
+ *
+ *  @date 2024/06/16
+ *  @author Evan Owens
+ *   */
+
+#ifndef INC_XBEE_H_
+#define INC_XBEE_H_
+
+#include <stdint.h>
+
+/*XBee API Packet Overhead*/
+#define START_DELIMITER 0x7E
+#define FRAME_TYPE 0x10
+#define FRAME_ID 0x01
+#define BIT_ADDRESS_16_LOW 0xFE
+#define BIT_ADDRESS_16_HIGH 0xFE
+#define BROADCAST_RADIUS 0x00
+#define OPTIONS 0x00
+#define TYPE 0x00
+#define BIT_ADDRESS_16_LOW 0xFE
+#define BIT_ADDRESS_16_HIGH 0xFE
+
+#define CHECKSUM_LENGTH_SIZE 4 //size of fields in api packet that are excluded in checksum and length field calculations
+
+/*XBee API Packet Positions*/
+#define START_DELIMITER_POSITION 0
+#define MSB_POSITION 1
+#define LSB_POSITION 2
+#define FRAME_TYPE_POSITION 3
+#define FRAME_ID_POSITION 4
+#define BIT_ADDRESS_64_START_POSITION 5
+#define BIT_ADDRESS_16_LOW_POSITION 13
+#define BIT_ADDRESS_16_HIGH_POSITION 14
+#define BROADCAST_RADIUS_POSITION 15
+#define OPTIONS_POSITION 16
+#define TYPE_POSITION 17
+#define LENGTH_POSITION 18
+#define MESSAGE_DATA_START_POSITION 19
+
+
+
+void XBEE_api_overhead_setup(uint8_t api_packet[]);
+void XBEE_calculate_checksum(uint8_t api_packet[], uint16_t packet_length);
+void XBEE_calculate_length(uint8_t api_packet[], uint16_t packet_length);
+
+#endif /* INC_XBEE_H_ */
--- a/components/tel/Core/Src/radio.c
+++ b/components/tel/Core/Src/radio.c
@@ -102,15 +102,15 @@ void RADIO_tx_API_Packager(uint8_t api_buffer[], uint16_t api_buffer_position)
 */
 void RADIO_tx_API_Accumulator(uint8_t radio_message[])
 {
-  int i;
-  for (i = 0; i<CAN_BUFFER_LEN; i++){
+
+  for (int i = 0; i<CAN_BUFFER_LEN; i++){
 	  api_buffer[i+api_buffer_position] = radio_message[i];
   }
 
   api_message_count++;
-  api_buffer_position = api_buffer_position + i;
+  api_buffer_position = api_buffer_position + CAN_BUFFER_LEN;
 
-  if (api_message_count == API_PACKET_SIZE){
+  if (api_message_count == API_MAX_MSGS){
 	  RADIO_tx_API_Packager(api_buffer, api_buffer_position);
 	  api_message_count = 0;
 	  api_buffer_position = 0;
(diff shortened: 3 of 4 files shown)


components/tel/Core/Inc/radio.h
```diff
@@ -43,6 +44,17 @@
 /* TIMING CONSTANTS */
 #define CAN_TRANSMIT_TIMEOUT                1000                // 1 second timeout
 
+/*XBee API Packet CONSTANTS*/
+#define API_BUFFER_SIZE 300 //number of bytes in api_buffeer array
+#define API_PACKET_SIZE 10 //number of messages in each api packet
+#define API_OVERHEAD_SIZE 18 //number of bits in api Overhead
```
AarjavJain101: Readability: Fix the tabbing and spacing to match what is already there.

components/tel/Core/Inc/xbee.h
```diff
@@ -0,0 +1,50 @@
...
+#define FRAME_TYPE_POSITION 3
+#define FRAME_ID_POSITION 4
+#define BIT_ADDRESS_64_START_POSITION 5
+#define BIT_ADDRESS_16_LOW_POSITION 13
+#define BIT_ADDRESS_16_HIGH_POSITION 14
+#define BROADCAST_RADIUS_POSITION 15
+#define OPTIONS_POSITION 16
+#define TYPE_POSITION 17
+#define LENGTH_POSITION 18
+#define MESSAGE_DATA_START_POSITION 19
```
AarjavJain101: Great constants makes it super easy to understand what you are doing.
Readability: Add tabbing and spacing a little ways from the define like its done in some other header files.

components/tel/Core/Src/radio.c
```diff
@@ -65,13 +73,59 @@ static void copy_CAN_id(uint8_t* source, CAN_Radio_msg_t *tx_CAN_msg, uint32_t c
...
+
+/**
+ * @brief Takes CAN radio buffer and adds it to an API buffer Array. Once enough messages are in the array, array is sent for packaging.
+ * @param radio_buffer[]: pointer to an array in which the messages are being added.
+ * @return void
+*/
+void RADIO_tx_API_Accumulator(uint8_t radio_message[])
+{
+  int i;
+  for (i = 0; i<CAN_BUFFER_LEN; i++){
```
AarjavJain101: Put the int i; initialization inside the for loop. This lets the reader know that the variable i will not be used later in the code but rather is used for iteration in the for loop.

components/tel/Core/Src/radio.c
```diff
@@ -65,13 +73,59 @@ static void copy_CAN_id(uint8_t* source, CAN_Radio_msg_t *tx_CAN_msg, uint32_t c
...
+	  api_buffer[i+api_buffer_position] = radio_message[i];
+  }
+
+  api_message_count++;
+  api_buffer_position = api_buffer_position + i;
```
AarjavJain101: I think you may have meant to also add the length of the CAN buffer. If you dont do this then your api_buffer_position will be 1 after you add a whole CAN message in which means the next CAN message that comes in will overwrite your previous CAN message starting at index 1.

components/tel/Core/Src/radio.c
```diff
@@ -65,13 +73,59 @@ static void copy_CAN_id(uint8_t* source, CAN_Radio_msg_t *tx_CAN_msg, uint32_t c
...
+
+  if (api_message_count == API_PACKET_SIZE){
```
AarjavJain101: Consider changing API_PACKET_SIZE to API_MAX_MSGS or something to indicate that this is the number of CAN messages in a Packet.

components/tel/Core/Src/radio.c
```diff
@@ -65,13 +73,59 @@ static void copy_CAN_id(uint8_t* source, CAN_Radio_msg_t *tx_CAN_msg, uint32_t c
...
+}
+
+
+
+/**
+ * @brief Takes CAN radio buffer and adds it to an API buffer Array. Once enough messages are in the array, array is sent for packaging.
+ * @param radio_buffer[]: pointer to an array in which the messages are being added.
+ * @return void
+*/
+void RADIO_tx_API_Accumulator(uint8_t radio_message[])
```
AarjavJain101, line 103: Maybe change accumulator function name to `RADIO_tx_API_Accumulate_and_Send` so that readers know that is where the UART call is made.

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * @brief Sets up overhead for an API Packet
+ *
+ * @param api packet: api packet array
+ * @return api_packet with setup api overhead
+*/
+void XBEE_api_overhead_setup(uint8_t api_packet[])
+{
+  uint8_t bit_address_64[] = {0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00};
```
AarjavJain101: Why is the bit_address_64 all 0's? Also I would make this a symbolic constant.

components/tel/Core/Src/radio.c
```diff
@@ -65,13 +73,59 @@ static void copy_CAN_id(uint8_t* source, CAN_Radio_msg_t *tx_CAN_msg, uint32_t c
...
+ * @return void
+*/
+void RADIO_tx_API_Packager(uint8_t api_buffer[], uint16_t api_buffer_position)
+{
+ uint16_t api_packet_size = API_OVERHEAD_SIZE + api_buffer_position;
+ uint8_t api_packet[api_packet_size];
+ XBEE_api_overhead_setup(api_packet);
+
+ for (int i = 0; i<api_buffer_position; i++){
+	 api_packet[MESSAGE_DATA_START_POSITION + i] = api_buffer[i];
```
AarjavJain101, line 87: Message data start position constant should be 18 not 19. This is because 18 bytes of overhead corresponds to index 17. So the first index of the first message would be the next byte which is index 18 (but byte 19).

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+*/
+void XBEE_calculate_length(uint8_t api_packet[], uint16_t packet_length)
+{
+ uint16_t total_bytes = packet_length - CHECKSUM_LENGTH_SIZE;
+ uint16_t lsb;
+ uint16_t msb;
+ lsb = total_bytes % 256;
+ msb = (uint16_t)(floor(total_bytes /256));
+ api_packet[MSB_POSITION] = msb;
+ api_packet[LSB_POSITION] = lsb;
```
AarjavJain101, line 75: Currently your MSB and LSB position are in an order that looks good for humans but if you want to read the length correclty on sunlink make sure you to api_packet at 1 is the MSB not the lower 8 bits of the length value (msb = upper 8 bits).

Just want to point this out so sunlink works accordingly.

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * @brief Calculates and adds the MSB and LSB for an API packet
+ *
+ * @param api packet: api packet array
+ * @return api_packet with setup length fields
+*/
+void XBEE_calculate_length(uint8_t api_packet[], uint16_t packet_length)
+{
+ uint16_t total_bytes = packet_length - CHECKSUM_LENGTH_SIZE;
```
AarjavJain101, line 69: Why do you do packet_length - CHECKSUM_LENGTH_SIZE? The packet length does not seem to include the checksum length in your code so it would be the equivalent of taking of CHECKSUM_LENGTH_SIZE bytes from the **original** packet.

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+ * @brief Adds up all the bytes after length and before checksum in an api packet
+ *
+ * @param api packet: api packet array
+ * @return uint16_t total_bytes in api packet before checksum and after length fields
+*/
+
+uint16_t XBEE_sum_bytes (uint8_t api_packet[], uint16_t packet_length)
+{
+  uint16_t total_bytes = 0;
+  for (int i = 3; i < packet_length; i++){
```
AarjavJain101: Add a symbolic constant here: `for (int i = 3; i < packet_length; i++){` for the number 3. Also this implies that the checksum is at the first 4 bytes of the API packet but I see in the overhead set up function that your first 4 bytes are delimeter, frame type, frame id, and bit address low.  I also thought that the checksum came at the end of the message (correct me on this)

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+
+  api_packet[START_DELIMITER_POSITION] = START_DELIMITER;
+  api_packet[FRAME_TYPE_POSITION] = FRAME_TYPE;
+  api_packet[FRAME_ID_POSITION] = FRAME_ID;
+  api_packet[BIT_ADDRESS_16_LOW_POSITION] = BIT_ADDRESS_16_LOW;
+  api_packet[BIT_ADDRESS_16_HIGH_POSITION] = BIT_ADDRESS_16_HIGH;
+  api_packet[BROADCAST_RADIUS_POSITION] = BROADCAST_RADIUS;
+  api_packet[OPTIONS_POSITION] = OPTIONS;
+
+  for (int i =0; i < 8; i++){
```
AarjavJain101: Add a symbolic constant for 8 here to indicate the length of the address in bytes

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+ * @brief Calculates the checksum of an api frame and adds it to the api_packet
+ *
+ * @param api packet: api packet array
+ * @return api_packet with setup checksum field
+*/
+void XBEE_calculate_checksum(uint8_t api_packet[], uint16_t packet_length)
+{
+ uint16_t total_bytes = XBEE_sum_bytes(api_packet, packet_length);
+ uint16_t checksum;
+ checksum = 255 - (total_bytes & 255);
```
AarjavJain101: Add a symbolic constant for 255 in checksum functions (Maybe like UINT8_MAX_SIZE). Also for readability change & 255 in the calculate checksum function to be & 0xFF (MASK_8_BITS).

components/tel/Core/Inc/radio.h
```diff
@@ -43,6 +44,17 @@
 /* TIMING CONSTANTS */
 #define CAN_TRANSMIT_TIMEOUT                1000                // 1 second timeout
 
+/*XBee API Packet CONSTANTS*/
+#define API_BUFFER_SIZE 300 //number of bytes in api_buffeer array
```
AarjavJain101: Currently, you will send extra null characters in the api_buffer because your api messages will be (10 CAN Messages) * (24 bytes per CAN message) + 18 bytes overhead = 258 bytes total. 300 - 258 = 42 bytes of null characters. This problem with this is that sunlink's process_message algorithm will take these null characters and assume they are a piece of the next api_message (since it could not cut it up into a valid message). Do your sunlink changes resolve this?

EvanO12: 300 is just the API buffer size. When I create the API packet array, it is based off the api_buffer position (which marks the end of can messages in the buffer) + the api overhead size.  This means, the api packet should be exactly the right size and no null characters will be transmitted. On sunlink side, if null characers were transmitted, we would get an error because the parser would try to parse groups of 24 null characters as a can message. This can be changed though using the length field (number of can messages) we transmit.

AarjavJain101: Right that makes sense. I see your code that sets the api_buffer to the value of the api_buffer_position. Thanks for clarifying!

components/tel/Core/Src/xbee.c
EvanO12: to calculate the checksum, you add up all the bytes from after the length so position 3 all the way to the end of the packet (before the checksum). So by doing what i do above, I am just ignoring the start delimiter, msb, and lsb. xctu or the api bom provides a more clear breakdown as my function does not necessarily add all the fields in order.

AarjavJain101: Makes sense. Thanks for clarification.

components/tel/Core/Src/xbee.c
EvanO12: Packet length should include the checksum (api_packet_size = API_OVERHEAD_SIZE + api_buffer_position) with API_OVERHEAD_SIZE including the checksum. What this is just doing is removing all the bytes from the start delimiter, length fields, and checksum so that length can be properly calculated.

components/tel/Core/Src/xbee.c
```diff
@@ -0,0 +1,87 @@
...
+  api_packet[START_DELIMITER_POSITION] = START_DELIMITER;
+  api_packet[FRAME_TYPE_POSITION] = FRAME_TYPE;
+  api_packet[FRAME_ID_POSITION] = FRAME_ID;
+  api_packet[BIT_ADDRESS_16_LOW_POSITION] = BIT_ADDRESS_16_LOW;
+  api_packet[BIT_ADDRESS_16_HIGH_POSITION] = BIT_ADDRESS_16_HIGH;
+  api_packet[BROADCAST_RADIUS_POSITION] = BROADCAST_RADIUS;
+  api_packet[OPTIONS_POSITION] = OPTIONS;
+
+  for (int i =0; i < 8; i++){
+	  api_packet[i] = bit_address_64[i];
```
AarjavJain101: I believe this line would overwrite your previous code that set the delimiter at index 0, frame type at index 3, and frame id at index 4?

components/tel/Core/Src/xbee.c
AarjavJain101: When you say API_OVERHEAD_SIZE including the checksum, where do you set the checksum in the api_buffer? Is this something that is not implemented yet or is it written to the API buffer elsewhere. The same applies for the length field. I ask this because i dont think I see it in the overhead_setup function here:
image

components/tel/Core/Src/radio.c
EvanO12: Good catch - I think I forgot to add the bytes of overhead that we add to the message (can byte and number of messages) into the overhead bytes count (and the rest of the code so I will have to implement that). Overall, there are 20 bytes of overhead. One byte, the checksum, comes after the message_data start position. So the data position should be 18.

components/tel/Core/Src/xbee.c
EvanO12: I wasn't sure how to make a symbolic constant of an array. We need to replace the 00s with the bit address of the radio receiver connected to sunlink

components/tel/Core/Src/xbee.c
EvanO12: right it should be api_packet[some symbolic constant i have defined + i] = 64address[i]
I wrote it that way, but had an error and removed the symbolic constant and forgot to add it back :(

components/tel/Core/Src/xbee.c
EvanO12: API_OVERHEAD_SIZE is a constant and its just the number of bytes of api overhead in every frame. Checksum is set by the XBEE_calculate_checksum(api_packet, api_packet_size); in the  RADIO_tx_API_Packager function

EvanO12: length is also set in the radio_tx_api_packager function by a function i wrote in the xbee.c file

components/tel/Core/Src/xbee.c
AarjavJain101: I see. For the symbolic constant you do them for each individual byte so long as its in a configurable in a header file.

components/tel/Core/Src/radio.c
```diff
@@ -81,6 +135,14 @@ void RADIO_tx_CAN_msg(CAN_Radio_msg_t *tx_CAN_msg)
   radio_buffer[CAN_CARRIAGE_RETURN_INDEX] = '\r';                                            /* CARRIAGE RETURN */
   radio_buffer[CAN_NEW_LINE_INDEX] = '\n';                                                   /* NEW LINE */
 
-  HAL_UART_Transmit(&huart1, radio_buffer, sizeof(radio_buffer), CAN_TRANSMIT_TIMEOUT);                     /* Transmit over UART */
-}   
+  RADIO_tx_API_Accumulator(radio_buffer);
```
ishanjoshi23, line 138: @AarjavJain101 Is working on the diagnostics API which will also introduce a UART API which can be used for sending out  radio messages. So the abstraction would look something like: XBee_API_Accumulator --> UART API --> Does uart transmit

components/tel/Core/Src/xbee.c
ishanjoshi23: @EvanO12 I would add this clarification in the function brief so anyone looking at the function knows exactly how the checksum is calculated.

components/tel/Core/Src/xbee.c
ishanjoshi23: same here, any relevant details should be mentioned in the function bride or as comments for readability.

components/tel/Core/Src/xbee.c
```diff
@@ -19,7 +20,7 @@
 uint16_t XBEE_sum_bytes (uint8_t api_packet[], uint16_t packet_length)
 {
   uint16_t total_bytes = 0;
-  for (int i = 3; i < packet_length; i++){
+  for (int i = FRAME_TYPE_POSITION; i < packet_length-2; i++){
```
AarjavJain101: What does the -2 signify?

components/tel/Core/Src/xbee.c
```diff
@@ -80,8 +87,8 @@ void XBEE_calculate_checksum(uint8_t api_packet[], uint16_t packet_length)
 {
  uint16_t total_bytes = XBEE_sum_bytes(api_packet, packet_length);
  uint16_t checksum;
- checksum = 255 - (total_bytes & 255);
- api_packet[packet_length] = checksum;
+ checksum = UINT8_MAX_SIZE - (total_bytes & MASK_8_BITS);
+ api_packet[packet_length-1] = checksum;
```
AarjavJain101: What does the -1 signify

PR #136, opened 2024-06-22
fix: Master works after adding GPS Delay + ExtId Fix

AarjavJain101:
**Problem**:
* master branch was not sending CAN messages.
* When CAN messages sent after fix EXTID was wrong.

**Fix:**
* CAN task was being starved because GPS task has NO delay when a fix is not acquired. That is why we saw one day that CAN messages + GPS was working together but never tested them in isolation. Thus a 10 second delay (same total delay as the fix = 1 case) was added to the fix = 0 case.
* Needed to change StdID variable to be dependent on whether we are sending ExtId or not.

**Testing before Merge**:
* Need to use this branch on the driving car and validate that it works. **DONE**
* **MUST VALIDATE THE FOLLOWING ON DRIVING CAR:** ~Correct GPS messages when a fix IS acquired~, no GPS or IMU messages when they are not connected, CAN messages always sending. Correct ExtId on Sunlink. **DONE**

### DIFF
--- a/components/tel/Core/Src/gps.c
+++ b/components/tel/Core/Src/gps.c
@@ -123,6 +123,7 @@ static void check_GPS_fix_for_send(GPS* gps_data) {
     }
     else
     {
+        osDelay(GPS_SINGLE_MSG_DELAY * 3 + GPS_NEXT_MSG_DELAY);              // 10000 ms (10 seconds) delay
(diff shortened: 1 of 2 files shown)


mjohal67 (approved):

ishanjoshi23 (approved):
Looks good.

PR #137, opened 2024-06-23
Fix: DID fixes

kcgroulx:
There was a issue where the mph was not being printed in the correct location and would wrap around.

IMG_4332

Release build of DID is not working at the moment. Need to do more root cause analysis. Since we're close to comp, we can use the Debug configuration for our testing. Added a #error when building with Release to prevent anyone from flashing the Release build accidently while we are debugging.

Added .bin as a post build output
image

- Updated gitignore for DID


ishanjoshi23:
I'm going to hold off on merging this until we try fixing the release build.

components/did/.settings/stm32cubeide.project.prefs
ubcsolar-admin, line 1: Did this need to be committed?

components/did/.settings/language.settings.xml
ubcsolar-admin, line 1: Did this need to be committed?

AarjavJain101, line 1: Did this need to be committed?

components/did/.settings/stm32cubeide.project.prefs
AarjavJain101, line 1: Did this need to be committed?

components/did/Core/Inc/LCD.h
```diff
@@ -39,13 +39,13 @@ Header file containing all constants and function declarations to the LCD
...
-// Target Cruise Speed
-#define CRUISE_XPOS 0
-#define CRUISE_YPOS 9
-#define CRUISE_DATA_XPOS 20
-#define CRUISE_DATA_YPOS 9
+// Regen Enabled
+#define REGEN_XPOS 0
+#define REGEN_YPOS 9
+#define REGEN_DATA_XPOS 20
+#define REGEN_DATA_YPOS 9
```
AarjavJain101: So we are replacing the Cruise display bit with regen now is what this looks like. Have we verified that this is accurately displayed on the Car? As in no wrapping AND correctly reading the incoming message?

components/did/Core/Src/DID.c
```diff
@@ -115,6 +116,14 @@ void parse_can_message( uint8_t* CAN_rx_data, uint32_t CAN_ID )
...
 			break;
 
+		/*
+		 *  MCB Diagnostics
+		 */
+		case CAN_ID_MCB_REGEN_ENABLED:
+			uint8_t regen_enabled = GETBIT(CAN_rx_data[3], 1);
+			SET_CYCLIC_DATA( data_MCB_regen_enabled, regen_enabled);
+			break;
+
```
AarjavJain101: The current MCB code shows that it is data byte 2 not 3?. If thats the case we should really make a define for this so that we dont remember the index we just remember what the bit represents.
image

components/did/Core/Src/DID.c
```diff
@@ -195,20 +204,20 @@ void update_DID_screen()
 			UpdateScreenTitles(PAGE_0);
 
 			// Get pointers to cyclic data variables
-			float* simulation_speed = GET_CYCLIC_DATA( data_simulation_speed );
+			float* vehicle_velocity = GET_CYCLIC_DATA( data_vehicle_velocity );
 			uint8_t* battery_soc = GET_CYCLIC_DATA( data_battery_SOC );
 			uint8_t* MCB_drive_state = GET_CYCLIC_DATA( data_MCB_drive_state );
-			float* target_velocity = GET_CYCLIC_DATA( data_target_velocity );
```
AarjavJain101, line 201: Since we dont get cyclic data for the target_velocity, we should remove the Setters and initializer for it. But also why are we removing it?
image
image

components/did/Core/Src/DID.c
```diff
@@ -250,22 +249,32 @@ void update_DID_screen()
...
+			if ( MCB_regen_enabled != NULL )
 			{
-				if ( *MCB_drive_state == 0x02 )
-				{
-					UpdateScreenParameter(CRUISE_DATA_XPOS, CRUISE_DATA_YPOS, (uint32_t)(*target_velocity), 0, FALSE);
-				}
-				else
-				{
-					OutputString("OFF", CRUISE_DATA_XPOS, CRUISE_DATA_YPOS);
+				if (*MCB_regen_enabled == 0x01) {
```
AarjavJain101: Lets make a define of 0x01.

components/did/Core/Src/DID.c
```diff
@@ -250,22 +249,32 @@ void update_DID_screen()
...
+					OutputString("ON", REGEN_DATA_XPOS, REGEN_DATA_YPOS);
+				} else {
+					OutputString("   ", REGEN_DATA_XPOS, REGEN_DATA_YPOS);
+					OutputString("OFF", REGEN_DATA_XPOS, REGEN_DATA_YPOS);
 				}
 			}
 			else
 			{
-				OutputString("---", CRUISE_DATA_XPOS, CRUISE_DATA_YPOS);
+				OutputString("---", REGEN_DATA_XPOS, REGEN_DATA_YPOS);
```
AarjavJain101, line 258: Consider adding a define for "---" incase we change what to display in the future

AarjavJain101 (changes requested):
Some points like the data byte for regen need to be confirmed or explained. Symbolic constants should be used as well.

components/did/Core/Src/DID.c
ishanjoshi23: You're right, it is the 3rd byte 0-indexed. I made a mistake - will fix it.

components/did/Core/Inc/LCD.h
ishanjoshi23: Have to validate on the car. We won't merge until it is validated.

components/did/Core/Src/DID.c
ishanjoshi23: This was cruise. Bad naming. There is a separate simulation velocity/target still there.

components/did/Core/Src/DID.c
ishanjoshi23: Sure

components/did/.settings/language.settings.xml
ishanjoshi23: I think we can add .settings: to the gitignore. Did @kcgroulx fix this in his other PR?

kcgroulx: fixed

components/did/.settings/stm32cubeide.project.prefs
kcgroulx: Fixed with merge from master

components/did/Core/Src/DID.c
ishanjoshi23: Done

components/did/Core/Src/DID.c
```diff
@@ -250,22 +252,53 @@ void update_DID_screen()
...
+//			}
+
+			// Pack Current
+			if ( pack_current != NULL )
+			{
+				UpdateScreenParameter(PACK_CURRENT_DATA_XPOS, PACK_CURRENT_DATA_YPOS, (uint32_t)(*pack_current), ((uint8_t)((*pack_current) * 10) % 10), TRUE);
+			}
+			else
+			{
+				OutputString("----", PACK_CURRENT_DATA_XPOS, PACK_CURRENT_DATA_YPOS);
```
AarjavJain101, line 280: Is this intended to be 4 dashes or 3 like the other ones. Want to make sure this wont cause something to be overwritten.

components/did/Core/Src/DID.c
```diff
@@ -250,22 +252,53 @@ void update_DID_screen()
...
+//			}
+//			else
+//			{
+//				OutputString("---", TARGET_DATA_XPOS, TARGET_DATA_YPOS);
+//			}
+
+			// Pack Current
+			if ( pack_current != NULL )
+			{
+				UpdateScreenParameter(PACK_CURRENT_DATA_XPOS, PACK_CURRENT_DATA_YPOS, (uint32_t)(*pack_current), ((uint8_t)((*pack_current) * 10) % 10), TRUE);
```
AarjavJain101, line 275: Why was the (packcurrent * 10) % 10 calculation chosen. Can we add a comment here for it.

ishanjoshi23: That gives us the decimal point.

components/did/Core/Src/DID.c
ishanjoshi23: I made it 4 dashes because the decimal point values are longer. I tested it with screen.

AarjavJain101 (approved):

ishanjoshi23:
IMG_3030 (1)

This is what the current screen looks like. The values are random using PCAN.

AarjavJain101 (approved):

ishanjoshi23 (approved):
Tested June 29

PR #138, opened 2024-06-26
Feat: MCB cylic data

kcgroulx:
## Pull Request Description

Added cyclic data library from DID to MCB. Used for vehicle velocity from MDI. If MDI does not send a vehicle velocity message after 1000ms. The MCB will trigger a MDI comm fault. MDI comm fault was also added to the MCB diagnostics message

## Monday Link

N/A

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [X] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [X] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

Tested on June 27th. Removed MDI connector. Verified MCB set MDI comm fault to true

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [X] Steps confirmed

## Sources

N/A


components/mcb/Core/Inc/mcb.h
```diff
@@ -61,9 +61,8 @@
  */
 #define ADC_LOWER_DEADZONE            900
 #define ADC_FOR_NO_SPIN               1300    
-#define ADC_MIN_FOR_FULL_THROTTLE     1830
-#define ADC_MAX_FOR_FULL_THROTTLE     2000
-
+#define ADC_MIN_FOR_FULL_THROTTLE     2000
+#define ADC_MAX_FOR_FULL_THROTTLE     2600
```
AarjavJain101, line 65: What experimentation led to these changes?

components/mcb/Core/Src/mcb.c
```diff
@@ -8,13 +8,19 @@
...
 #include "mcb.h"
 #include "can.h"
+#include "cyclic_data.h"
 
 // githash.h is autogenerated as a prebuild step.
 // If you are getting build errors here, manually create a githash.h file in the same directory and add #define GITHASH "********"
 #include "../../../../libraries/githash/githash.h" // TODO: Add this library to the project instead of using the relative path
 
+
+#define MAX_CYCLE_TIME_VEHICLE_VELOCITY 1000 // Max time(ms) of not receive vehicle velocity from MDI before setting a flaut in the mcb diagnosics MCB message
```
AarjavJain101: small thing: fault spelled wrong.

components/mcb/Core/Src/mcb.c
```diff
@@ -189,7 +200,20 @@ void UpdateInputFlags(InputFlags * input_flags)
...
+	if (velocity != NULL)
+	{
+		input_flags->velocity_under_threshold = *velocity < VELOCITY_THRESHOLD;
+		input_flags->MDI_communication_fault = false;
+	}
+	else
+	{
+		// If NULL, then we haven't received a vehicle velocity msg from the MDI for MAX_CYCLE_TIME_VEHICLE_VELOCITY.
+		// Set 
+		input_flags->velocity_under_threshold = false;
```
AarjavJain101, line 214: This might already be tested but I will ask to make sure: By setting velocity under threshold to false in the error state (MDI gives no message) this would mean that the firmware believes the car speed > 0.5 m/s. So if you wanted to go into park/transition to any state you would not be able to because velocity under threshold is false. This would be incorrect behavior in cases where we notice a fault and want to park the car or to reverse into a place where we can stop and diagnose the issue correct? Or is velocity_under_threshold being set elsewhere so its not permanently false?

AarjavJain101 (commented):
Approving once `velocity_under_threshold` discussion complete.

components/mcb/Core/Inc/mcb.h
kcgroulx: On Thursday, we found that the ADC out of range fault would be triggered when Diego pressed the pedal all the way. We recalibrated theses values using the mcb diagnostics message.

AarjavJain101: Got it. Thanks

components/mcb/Core/Src/mcb.c
kcgroulx: Since we don't have velocity measurements, we can't know for certain if the car is slow enough to change drive states.
In the future, we can add a new FAULT state that the MCB will enter when a fault is detected. Behavior in the FAULT state would need to be discussed.

AarjavJain101: Since we already set the MDI no velocity fault to true could we use that as the criteria for the FAULT state? I agree it should be discussed further with some others. I think that since we simply cant know the velocity and the MCB cant override or do any motor controlling like the MDI does I would say ya we cant switch states and instead we just have to hold the brake.

ishanjoshi23 (approved):

ishanjoshi23:
Tested on the car on Thursday June 27

AarjavJain101 (approved):

PR #142, opened 2024-06-30
Fix: DID Pack Current Signed Value + Release build

kcgroulx:
## Pull Request Description

Draft: Needs vehicle testing
Updating Pack Current to be a Signed Value
- Negative value means that it is charging.
- Positive means current out of the pack.

This fix is necessitated due to Mischa noticing a very large value when charging the battery.

## Effected Components

- [ ] AMB
- [ ] BMS
- [x] DID
- [ ] ECU
- [ ] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [x] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


ishanjoshi23 (approved):
Need to test. Don't merge until tested on car.

ishanjoshi23:
Flashed onto the DID. Could not test with a negative value though due to not charging right now. We have to just check when we charge next.

ishanjoshi23:
Release build is still not working

ishanjoshi23:
These values did not look correct (negative) when charging. We have to re-do this.

PR #144, opened 2024-07-03
[Draft] Feat: MCB timeout for MDI

kcgroulx:
## Pull Request Description

Added mcb communication fault if mdi does not receive a motor command from the mcb in >500ms.  
mdi will send `0.0f` acceleration and `0.0f` regen to mdu if mcb comm fault is triggered

## Monday Link

N/A

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [ ] MCB
- [X] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

needs testing

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed

### DIFF
--- a/components/mdi/Core/Inc/can.h
+++ b/components/mdi/Core/Inc/can.h
@@ -46,6 +46,7 @@ typedef struct
 	bool PWR_mode_on;
 	bool regen;
 	bool cruise_control_enable;
+	uint32_t last_rx_time;
 
 	//0x501
 	uint8_t motorCurrentFlag;
--- a/components/mdi/Core/Src/can.c
+++ b/components/mdi/Core/Src/can.c
@@ -19,6 +19,8 @@
 */
 void CAN_Decode_Velocity_Message(uint8_t localRxData[], CAN_message_t* CAN_msg)
 {
+	// Update the last rx time
+	CAN_msg->last_rx_time = HAL_GetTick();
 
 	union { //union struct to convert the input stream of 32-bits into IEEE float
 		uint32_t concatenated_bits;
--- a/components/mdi/Core/Src/main.c
+++ b/components/mdi/Core/Src/main.c
@@ -329,7 +329,9 @@ int main(void)
         TxHeader.StdId = MDI_DIAGNOSTICS_CAN_ID;
         TxHeader.IDE = CAN_ID_STD;
         TxHeader.DLC = 1;
-        TxData[0] = 0; // Clear TxData[0]
+        for(int i = 0; i < 8; i++){   // Clear TxData
+          TxData[i] = 0;
+        }
 
         if(1 == mcb_communication_fault)
         {
--- a/components/mdi/Core/Src/main.c
+++ b/components/mdi/Core/Src/main.c
@@ -40,6 +40,8 @@
 /* USER CODE BEGIN PM */
 #define DAC_ACC_ADDR    (0b0001110 << 1) //I2C address for the acceleration DAC
 #define DAC_REGEN_ADDR  (0b0001101 << 1) //I2C address for the regen DAC
+
+#define MCB_COMMUNICATION_TIMEOUT 500 // Time in ms to wait for a motor command from MCB before setting a fault
 /* USER CODE END PM */
 
 /* Private variables ---------------------------------------------------------*/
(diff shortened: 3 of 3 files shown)


components/mdi/Core/Inc/can.h
```diff
@@ -46,6 +46,7 @@ typedef struct
 	bool PWR_mode_on;
 	bool regen;
 	bool cruise_control_enable;
+	uint32_t last_rx_time;
```
ishanjoshi23, line 49: We talked about moving this to a global variable of its own; this introduces a lot of changes that need to be made in the MDI firmware (after comp). This is good for now.

components/mdi/Core/Src/can.c
```diff
@@ -19,6 +19,8 @@
 */
 void CAN_Decode_Velocity_Message(uint8_t localRxData[], CAN_message_t* CAN_msg)
 {
+	// Update the last rx time
+	CAN_msg->last_rx_time = HAL_GetTick();
```
ishanjoshi23, line 23: Basically last time is ONLY updated for the 0x401 message, and no other

components/mdi/Core/Src/main.c
```diff
@@ -211,7 +215,21 @@ int main(void)
     //////////////////////////////////////
     // SEND SIGNALS TO MOTOR CONTROLLER //
     //////////////////////////////////////
-    if( (send_data_flag == 1 ) && (Parse_Data_Flag == 0) )
+
+    // Check for mcb timeout
+    if(HAL_GetTick() > msg0.last_rx_time + MCB_COMMUNICATION_TIMEOUT)
```
ishanjoshi23, line 223: Test my disconnecting MCB CAN cable

components/mdi/Core/Src/main.c
```diff
@@ -211,7 +215,21 @@ int main(void)
...
+    {
+      if(0 == mcb_communication_fault)
+      {
+        // Send 0.0f acceleration
+        Send_Voltage(0.0f, DAC_ACC_ADDR, &hi2c2);
+        // Send 0.0f regen
+        Send_Voltage(0.0f, DAC_REGEN_ADDR, &hi2c2);
```
ishanjoshi23, line 230: Checked this in an onine compiler. Floating point 0 * 1024 typecasted to uint16_t returns 0.

components/mdi/Core/Src/main.c
```diff
@@ -211,7 +215,21 @@ int main(void)
...
+
+        mcb_communication_fault = 1;
+      }
+    }
+    else if( (send_data_flag == 1 ) && (Parse_Data_Flag == 0) )
```
ishanjoshi23, line 235: Only if no MCB timeout - good

components/mdi/Core/Src/main.c
```diff
@@ -211,7 +215,21 @@ int main(void)
...
+    if(HAL_GetTick() > msg0.last_rx_time + MCB_COMMUNICATION_TIMEOUT)
+    {
+      if(0 == mcb_communication_fault)
+      {
+        // Send 0.0f acceleration
+        Send_Voltage(0.0f, DAC_ACC_ADDR, &hi2c2);
+        // Send 0.0f regen
+        Send_Voltage(0.0f, DAC_REGEN_ADDR, &hi2c2);
+
+        mcb_communication_fault = 1;
```
ishanjoshi23, line 232: Ensures that the 0, 0 command is only sent once per timeout.

ishanjoshi23 (approved):

components/mdi/Core/Src/main.c
```diff
@@ -303,6 +324,34 @@ int main(void)
     		}
     		HAL_CAN_AddTxMessage(&hcan, &TxHeader, TxData, TxMailbox);
     		SendSlowMessage = 0;
+
+        // MDI diagnostics 0x50F
+        TxHeader.StdId = MDI_DIAGNOSTICS_CAN_ID;
+        TxHeader.IDE = CAN_ID_STD;
+        TxHeader.DLC = 1;
+        TxData[0] = 0; // Clear TxData[0]
```
ishanjoshi23: The entire array should be cleared here, not just byte 0 (the rest will be garbage data)

PR #150, opened 2024-07-14
Backstop added to pedal so adjusting the max value

ishanjoshi23:
## Pull Request Description

Backstop was added to the accel pedal during comp. Adjusting the max ADC reading here for full throttle. Below is images of the data we got for the full and no pedal press.

IMG_3361-min
IMG_3362-min

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [x] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [x] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed

### DIFF
--- a/components/mcb/Core/Inc/mcb.h
+++ b/components/mcb/Core/Inc/mcb.h
@@ -61,7 +61,7 @@
  */
 #define ADC_LOWER_DEADZONE            900
 #define ADC_FOR_NO_SPIN               1300    
-#define ADC_MIN_FOR_FULL_THROTTLE     2000
+#define ADC_MIN_FOR_FULL_THROTTLE     1850
 #define ADC_MAX_FOR_FULL_THROTTLE     2600
 
 #define SETBIT(x, bitpos) (x |= (1 << bitpos))


components/mcb/Core/Inc/mcb.h
```diff
@@ -61,7 +61,7 @@
  */
 #define ADC_LOWER_DEADZONE            900
 #define ADC_FOR_NO_SPIN               1300    
-#define ADC_MIN_FOR_FULL_THROTTLE     2000
+#define ADC_MIN_FOR_FULL_THROTTLE     1850
```
AarjavJain101, line 64: 1850 makes sense based on the data you provided.

components/mcb/Core/Inc/mcb.h
```diff
@@ -61,7 +61,7 @@
  */
 #define ADC_LOWER_DEADZONE            900
 #define ADC_FOR_NO_SPIN               1300
```
AarjavJain101, line 63: Based on the data you showed I would go for 1350-1400. The reason being is that the lower line that your data showed **I am assuming** means they were completely off the pedal. Therefore having `ADC_FOR_NO_SPIN` be 1330 would mean that the pedal just sitting there could result in acceleration because of noise. **We want a threshold outside of this noisy region**. In your picture it seems that you chose the lowest point of the noise so I would actually take the highest point of the noise and add 20 to it and make that the `ADC_FOR_NO_SPIN` value. As a guess based on what you showed something like 1375 might be good.

AarjavJain101 (changes requested):

kcgroulx (approved):

PR #151, opened 2024-07-15
Fix: Competition current threshold modification

mjohal67:
## Pull Request Description
Regulation 8.3.B states:
> All lithium based battery packs must have active protection such that over-voltage,
over-temperature (for charge and discharge rating), over-current and under-voltage cause the Main
Power Switch per 8.6.A to open and to electrically isolate the source or sink for the vehicle.

The "OR" statement at the end of this regulation caused us to have issues at scruitineering. We assumed that the over-temperature charge limit would be a trip, and that the over-temperature discharge limit would be a fault. However, the scruitineers required us to have a fault at 45C (trip at 44C), _and_ a fault at 65C (whereas before, we just had the trip at 45C and fault at 65C). 

They argue that this because, when you exceed any of the manufactuer's specifications, you should fault the pack. Thus, even though you're opening HLIM when you trip at 44C and requesting regen off (meaning you'd never reach 45C while charging), you still need to trip at 45C. 

BMS FW was changed to trip at 44C, and a fault was added at 45C with an additional check on the current direction of the ECU current (which we get from the ECU's CAN current field). The 65C fault was unchanged. There was no additional CAN bit that was set for this. 

See updated ECU status code ref: https://docs.google.com/spreadsheets/d/1xxisHuuusQCDnes3OIaucntGGlHj25iUrbfAIXvNhaM/edit?gid=0#gid=0

## Effected Components

- [ ] AMB
- [x] BMS
- [ ] DID
- [ ] ECU
- [ ] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [x] Vehicle testing
- [x] Bench top testing
- [ ] Unit testing
- [x] Other (Comp 2024 scruitineering) 
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


ishanjoshi23 (approved):

Tik-Hakobyan (approved):

ishanjoshi23 (changes requested):
Remove AMB changes

ishanjoshi23 (approved):
I fixed the extra files added here

mjohal67:
Related Monday update: https://ubcsolar26.monday.com/boards/7524358047/pulses/7524360025/posts/4176716691

PR #153, opened 2024-08-01
Fix: ECU-BMS FLT GPIO Checking for Array Faulting

mjohal67:
## Pull Request Description

When arrays were plugged into the battery, at high irradiances (i.e. currents), the pack would fault, for no apparent reason (no fault statuses lit up on telemetry). The only place where a fault could get triggered without it showing up in the fault code is with noise being induced on one of the fault GPIOs from the masterboard to the ECU.

Prior to this change, there were three GPIOs that the ECU would read and, if set, would transition to a fault state on. FLT, COMM, and OT. 

**Firmware Changes**:
* ECU no longer cares about COMM and OT GPIOs. These are encapsulated within the FLT GPIO from the masterboard's perspective, so they're redundant information. 
* Scrutineer Ryan suggested doing a "checksum" using a spare GPIO and the FLT GPIO. I used the BAL GPIO from the masterboard, and set it high at the beginning of the BMS's firmware.

When the BMS starts up, it'll set BAL high, indicating no fault, and will set FLT high (indicating fault). As normal, the BMS will check the pack once and set the fault pin accordingly, signalling to the ECU to continue with its startup sequence (or to go to a fault state).

Note: During startup, the ECU doesn't care about the BAL pin (no need, since arrays won't be opened until later in startup). We also never saw issues with the arrays causing the battery to fault on startup. 

During normal monitoring, if the BMS faults, it'll set FLT high and BAL low - if the ECU reads that _combination_ of GPIOs, only then will it fault. So, if either FLT goes high, or BAL goes low individually, the ECU won't fault (hence, checksum). 

This fix was implemented prior to morning charging of day 3 of competition, and was tested in front of scrutineers by pulling out the slave board connection to the masterboard, inducing communications fault. For the entirety of day 3 of race, we saw no faults (the sunniest day of FSGP 2024) - additionally, there were no faults during ASC. I believe it's safe to say this fixed the issue.

## Effected Components

- [ ] AMB
- [x] BMS
- [ ] DID
- [x] ECU
- [ ] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD

## Testing

- [x] Vehicle testing
- [x] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [x] Steps confirmed


components/bms/bms_master/Core/Src/bms_main.c
```diff
@@ -221,16 +221,17 @@ void BMS_MAIN_driveOutputs(Pack_t *pack)
...
         CONT_OT_switch(pack->status.bits.fault_over_temperature);
         stopBalancing(pack);
-        CONT_BAL_switch(pack->status.bits.balancing_active);
         CONT_FAN_PWM_set(FAN_FULL);
     }
     else // no fault; balance modules, drive control signals and fans
     {
         BAL_updateBalancing(pack); // write bal settings, send bal commands
-        CONT_BAL_switch(pack->status.bits.balancing_active);
+        // CONT_BAL_switch(pack->status.bits.balancing_active);
```
ishanjoshi23, line 234: I just want to make sure here that the intention **is** to comment this out?

ishanjoshi23 (approved):
Everything looks ok and agrees with what you wrote in the PR description. Just wanted to double check with you @mjohal67 that the commented out line is indented to be commented out. Has this been documented in Monday as well?

components/bms/bms_master/Core/Src/bms_main.c
mjohal67: Correct.

Tik-Hakobyan (approved):

PR #160, opened 2024-10-21
Feat: CAN_comm library

kcgroulx:
## Pull Request Description

Implemented library for CAN_comms. 

## Monday Link

https://ubcsolar26.monday.com/boards/7524367653/pulses/7595111315

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [ ] MCB
- [ ] MDI
- [ ] TEL
- [ ] CI/CD
- [X] Library

## Testing

- [ ] Vehicle testing
- [X] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

Implemented library on TEL on a separate branch. Used PCAN to verify messages can be transmitted and received.

image

image

image

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [X] Steps confirmed


libraries/CAN_comms/Src/CAN_comms.c
```diff
@@ -0,0 +1,287 @@
...
+CAN_comms_config_t CAN_comms_config;
+osMessageQueueId_t CAN_comms_Rx_queue;
+osMessageQueueId_t CAN_comms_Tx_queue;
+osThreadId_t CAN_comms_Rx_task_handle;
+osThreadId_t CAN_comms_Tx_task_handle;
+osSemaphoreId_t CAN_comms_Tx_mailbox_semaphore;
+const osThreadAttr_t CAN_comms_Rx_task_attributes = {
+    .name = "CAN_comms_Rx_task",
+    .stack_size = CAN_RX_TASK_STACK_SIZE,
+    .priority = (osPriority_t) osPriorityLow,
```
EvanO12, line 37: Why osPriorityLow vs high or normal?

libraries/CAN_comms/Src/CAN_comms.c
```diff
@@ -0,0 +1,287 @@
...
+}
+
+
+/**
+ * @brief Interrupt Service Routine for the CAN mailbox callback function
+ * This function frees the semaphore for the CAN Tx mailbox
+ * 
+ * @attention This function needs to be added to each CAN mailbox complete callback function
+ */
+void CAN_comms_Tx_mailbox_complete_ISR()
```
EvanO12, line 218: Where is this function called? I see each mailbox call their own semaphore release.

EvanO12 (approved):

libraries/CAN_comms/Src/CAN_comms.c
kcgroulx: Good catch, I'll remove this function.

libraries/CAN_comms/Src/CAN_comms.c
AarjavJain101: Good point. @kcgroulx I think this should be high.

libraries/CAN_comms/Src/CAN_comms.c
```diff
@@ -0,0 +1,287 @@
...
+osSemaphoreId_t CAN_comms_Tx_mailbox_semaphore;
+const osThreadAttr_t CAN_comms_Rx_task_attributes = {
+    .name = "CAN_comms_Rx_task",
+    .stack_size = CAN_RX_TASK_STACK_SIZE,
+    .priority = (osPriority_t) osPriorityLow,
+};
+const osThreadAttr_t CAN_comms_Tx_task_attributes = {
+    .name = "CAN_comms_Tx_task",
+    .stack_size = CAN_TX_TASK_STACK_SIZE,
+    .priority = (osPriority_t) osPriorityLow,
```
AarjavJain101: Same with this task.

libraries/CAN_comms/Inc/CAN_comms.h
```diff
@@ -0,0 +1,59 @@
...
+ * 
+ * @attention See README for details on how to implement this library.
+ */
+
+#ifndef CAN_COMMS_H
+#define CAN_COMMS_H
+
+#include <stdint.h>
+
+#include "can.h"
```
AarjavJain101: Is including can.h necessary? Or can we remove it? I wanted to declare the callback inside of can.h but cant because CAN comms.h will include can.h meaning it will try to find the defintion a typedef you have defined in CAN_comms.h.

libraries/CAN_comms/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+**Receiving**
+
+- CAN Rx messages are stored in memory as a `CAN_comms_Rx_msg_t` struct. Which contains a `CAN_RxHeaderTypeDef` and a `uint8_t` array parameters.
+- The `CAN_comms_Rx_callback` function pointer is called from the `CAN_comms_Rx_task` whenever a CAN message is received. This is where the user impliments functionality for receiving CAN messages.
+
+  ```c
+  void CAN_comms_Rx_callback(CAN_comms_Rx_msg_t* CAN_comms_Rx_msg)
+  {
+    // User implimention for parsing CAN_comms_Rx_msg
```
AarjavJain101: miss spell implementation lol

libraries/CAN_comms/README.md
```diff
@@ -0,0 +1,99 @@
...
+  CAN_comms_config_t CAN_comms_config = {
+      .hcan = &hcan,
+      .CAN_Filter = CAN_filter,
+      .CAN_comms_Rx_callback = CAN_comms_Rx_callback
+    };
+  ```
+
+- `hcan` should be a pointer to the `CAN_HandleTypeDef` struct that's normally defined in `can.h`
+- `CAN_Filter` should be a `CAN_FilterTypeDef` struct.
+- `CAN_comms_Rx_callback` should be a pointer to a function of the form `void (*CAN_comms_Rx_callback)(CAN_comms_Rx_msg_t*);` This function will be called whenever a CAN message is received and ready to be parsed.
```
AarjavJain101: Mention where to define this function. If you remove can.h then defining this callback in can.h is probably the most reasonable spot.

libraries/CAN_comms/Src/CAN_comms.c
kcgroulx: There is an edge case where the TEL board receives so many messages to the point where the CAN_comms_Rx_task is not able to empty the queue fast enough. If this task has a higher priority than other tasks, then it will stay in the CAN_comms_Rx_task and not switch. This causes an issue if other tasks are doing important things like refreshing the watchdog. That's why I chose to use osPriorityLow.

libraries/CAN_comms/Inc/CAN_comms.h
kcgroulx: CAN_comms requires some declarations and typedefs that are included in can.h. However, I agree that defining the Rx callback in can.h makes the most sense. Let me do some digging to see if there is a way to resolve this.

libraries/CAN_comms/Src/CAN_comms.c
kcgroulx: I'll change it to osPriorityHigh.

libraries/CAN_comms/Inc/CAN_comms.h
kcgroulx: Removed can.h and it works

AarjavJain101 (approved):

PR #161, opened 2024-10-25
Feat: CAN Busload TEL

makac1896:
## Pull Request Description

This feature focuses on analyzing our CAN bus and reporting the overall usage of the network for better understanding and debugging. CAN bus load is a measure of how much of our CAN bus we are using and in this PR, the feature is implemented, including the calculations and the new CAN message sent out to other boards that report the bus load.

## Monday Link
https://ubcsolar26.monday.com/boards/7524367653/pulses/7524368220

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [ ] MCB
- [ ] MDI
- [x] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [x] Unit testing
- [x] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [x] Steps confirmed

## Sources

All sources and related documentation are available on the linked Monday card. I am working on a unified resource that will be updated on the wiki.


components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,72 @@
+#include "canbusload.h"
+
+uint32_t can_total_bits = 0;
+static uint64_t counter_window = 0;
+static uint64_t current_bus_load = 0;
+uint32_t currentIdx = 0;
+uint32_t circularBuffer[CANLOAD_WINDOW_SIZE] = {0};
+
+uint32_t CANLOAD_calculate_message_bits(CAN_msg_t *msg)
```
makac1896: This is another public function to be included in the TEL interrupt when new messages are received and it keeps a running total of all bits sent over the specified windwo

components/tel/Core/Inc/canload.h
```diff
@@ -0,0 +1,39 @@
...
+#define CANLOAD_RTR_BITS 1
+#define CANLOAD_IDE_BITS 1
+#define CANLOAD_RESERVED_BIT_R0 1
+#define CANLOAD_DLC_BITS 4
+#define CANLOAD_CRC_BITS 15
+#define CANLOAD_CRC_DELIMITER_BITS 1
+#define CANLOAD_ACK_SLOT_BITS 1
+#define CANLOAD_ACK_DELIMITER_BITS 1
+#define CANLOAD_EOF_BITS 7
+#define CANLOAD_IFS_BITS 3
```
AarjavJain101: No need for the CANLOAD prefix because these will not be used by others (only your private .c source code needs this defines). Additionally, since these will not be used by the user then these defines should go into a private define section of you .c source code

See the following reference to the Coding Style on the Wiki
image

components/tel/Core/Src/canload.c
AarjavJain101: Instead of standardizing the input of the function to be a struct you made, lets pass in just the header of the CAN message which has a type **already defined in the STM32 HAL**. Thoughts @kcgroulx

AarjavJain101: Also as of current `CAN_msg_t` is only something that exists in the old TEL code so currently you would need to create that typedef in this code.

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,72 @@
...
+    bits += CANLOAD_CRC_BITS;
+    bits += CANLOAD_CRC_DELIMITER_BITS;
+    bits += CANLOAD_ACK_SLOT_BITS;      
+    bits += CANLOAD_ACK_DELIMITER_BITS;
+    bits += CANLOAD_EOF_BITS;
+    bits += CANLOAD_IFS_BITS;
+    return bits;
+}
+
+void update_sliding_window(uint32_t bits)
```
AarjavJain101: Add function header. See the function header example of the Coding Style guide on the Wiki
image

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,72 @@
...
+
+float calculate_bus_load()
+{
+    return ((float) calculate_total_bits() / ((float)CANLOAD_WINDOW_SIZE * (float)CANLOAD_BIT_RATE)) * 100.0;
+}
+
+float CANLOAD_get_bus_load()
+{    
+	if(counter_window % 10 == 0){
+        // Runs every 1 second
```
AarjavJain101: Change the comment to explain where 1 second was calculated

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,72 @@
...
+}
+
+float calculate_bus_load()
+{
+    return ((float) calculate_total_bits() / ((float)CANLOAD_WINDOW_SIZE * (float)CANLOAD_BIT_RATE)) * 100.0;
+}
+
+float CANLOAD_get_bus_load()
+{    
+	if(counter_window % 10 == 0){
```
AarjavJain101: The 10 here should be defined in this .c file

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,72 @@
...
+        // Runs every 1 second
+		  update_sliding_window(can_total_bits);
+	}
+	if(counter_window % 50 == 0){
```
AarjavJain101: The 50 here should be defined in this .c file

components/tel/Core/Inc/canload.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * canbusload.h
+ *
+ *  Created on: Oct 17, 2024
+ *      Author: makac
```
AarjavJain101: Remove the author + date part

components/tel/Core/Inc/canload.h
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+#ifndef INC_CANBUSLOAD_H_
+#define INC_CANBUSLOAD_H_
+
+#include "can.h"
+
+#define CANLOAD_BIT_RATE 500000
```
AarjavJain101: Add a link in the comments to the monday update you wrote that explains where all these constants came from

components/tel/Core/Src/canload.c
kcgroulx: You can just pass in the DLC as an arg instead of the entire struct

components/tel/Core/Inc/canload.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * canbusload.h
+ *
+ *  Created on: Oct 17, 2024
+ *      Author: makac
+ */
+
```
kcgroulx: We don't need to add the date and author here, this would be tracked by git.

components/tel/Core/Inc/canload.h
kcgroulx: The baudrate might already be defined somewhere

components/tel/Core/Inc/canload.h
kcgroulx: Are these defines used outside of canload.c? If not, I would move these defines to the .c instead.

components/tel/Core/Inc/canload.h
```diff
@@ -0,0 +1,39 @@
...
+#define CANLOAD_RESERVED_BIT_R0 1
+#define CANLOAD_DLC_BITS 4
+#define CANLOAD_CRC_BITS 15
+#define CANLOAD_CRC_DELIMITER_BITS 1
+#define CANLOAD_ACK_SLOT_BITS 1
+#define CANLOAD_ACK_DELIMITER_BITS 1
+#define CANLOAD_EOF_BITS 7
+#define CANLOAD_IFS_BITS 3
+
+extern uint32_t can_total_bits;
```
kcgroulx: Does this need to be extern?

kcgroulx (changes requested):

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,142 @@
...
+#define DLC_BITS 4
+#define CRC_BITS 15
+#define CRC_DELIMITER_BITS 1
+#define ACK_SLOT_BITS 1
+#define ACK_DELIMITER_BITS 1
+#define EOF_BITS 7
+#define IFS_BITS 3
+
+static uint32_t can_total_bits = 0;
+static uint64_t current_bus_load = 0;
```
AarjavJain101: Small thing: ask yourself if 8 bytes is needed to store the bus load in any case. If its not I would change this to the smallest value that is still large enough for the worst case.

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,142 @@
...
+ * This function modifies the global variable `can_total_bits`. It does not
+ * return any value.
+ *
+ * @param DLC The data length code of the CAN message.
+ */
+void CANLOAD_calculate_message_bits(uint32_t DLC)
+{
+    uint32_t bits = 0;
+    bits += SOF_BITS;
+    if (DLC == 0)
```
AarjavJain101: Why does the DLC equalling 0 mean that there is a standard ID present?

components/tel/Core/Src/canload.c
```diff
@@ -0,0 +1,142 @@
...
+ * This function calculates the number of bits in a CAN message based on the
+ * data length code (DLC) of the message. The result is stored in the global
+ * variable `can_total_bits`.
+ *
+ * This function modifies the global variable `can_total_bits`. It does not
+ * return any value.
+ *
+ * @param DLC The data length code of the CAN message.
+ */
+void CANLOAD_calculate_message_bits(uint32_t DLC)
```
AarjavJain101: Instead of exposing this function to the user, is it possible to just make this function private, and call it first thing in the `CANLOAD_update_sliding_window` function? You would have to add the CAN ID and DLC as an input though.

components/tel/Core/Inc/canload.h
```diff
@@ -0,0 +1,12 @@
+#ifndef INC_CANBUSLOAD_H_
+#define INC_CANBUSLOAD_H_
+
+#include "can.h"
+
+#define CANLOAD_BAUD_RATE 500000
```
AarjavJain101: Does this need to be exposed in `can.h`

AarjavJain101:
@makac1896: Can you change the name of the PR to follow the standard naming convention. See [this guide](https://wiki.ubcsolar.com/en/subteams/embedded/git-usage) as well as older PR names.

components/tel/Core/Src/freertos.c
```diff
@@ -195,6 +207,26 @@ void GPS_task(void *argument)
...
+/* USER CODE END Header_CANBusLoad_Task */
+void CANBusLoad_Task(void *argument)
+{
+  /* USER CODE BEGIN CANBusLoad_Task */
+  /* Infinite loop */
+  for(;;)
+  {
+    CANLOAD_update_sliding_window();
+    CAN_tx_canload_msg();
+    osDelay(1);
```
AarjavJain101: Change the timing of the logic instead of sending this message every 1 millisecond.

components/tel/Core/Src/can.c
```diff
@@ -169,4 +182,11 @@ void HAL_CAN_ErrorCallback(CAN_HandleTypeDef *hcan)
     }
 }
 
+void CAN_tx_canload_msg() {
```
AarjavJain101, line 242: Add function header

components/tel/Core/Src/freertos.c
```diff
@@ -195,6 +207,26 @@ void GPS_task(void *argument)
...
+* @retval None
+*/
+/* USER CODE END Header_CANBusLoad_Task */
+void CANBusLoad_Task(void *argument)
+{
+  /* USER CODE BEGIN CANBusLoad_Task */
+  /* Infinite loop */
+  for(;;)
+  {
+    CANLOAD_update_sliding_window();
```
AarjavJain101: Since your task sends the CAN bus load message, it takes up room on the CAN bus. Your current implementation does not add the bits of the CAN bus load message itself. To fix this, do a call do `CANLOAD_calculate_message_bits()` right before you update the sliding window.

components/tel/.cproject
AarjavJain101, line 1: @kcgroulx Does the .cproject need to be committed?

components/tel/Core/Src/usart.c
```diff
@@ -49,7 +49,7 @@ void MX_USART1_UART_Init(void)
 
   /* USER CODE END USART1_Init 1 */
   huart1.Instance = USART1;
-  huart1.Init.BaudRate = 230400;
+  huart1.Init.BaudRate = 921600;
```
AarjavJain101: You need to re-merge master because master shows 230400.
image

tools/CAN_Testing/BS25_CAN_ID_Table_Simulate.xmt
```diff
@@ -0,0 +1,61 @@
+; C:\Users\makac\git\firmware_v3\tools\CAN_Testing\BS25_CAN_ID_Table_Simulate.xmt
+;   CAN messages saved by PCAN-View v5.1.1.908
+;
+; Columns descriptions:
+; ~~~~~~~~~~~~~~~~~~~~~
```
AarjavJain101, line 5: I am not sure if this was your intention @makac1896 but all the messages are paused. But I actually think thats a good thing because then people can choose what messages they want to turn on. @kcgroulx what do you think?

components/tel/Core/Src/freertos.c
```diff
@@ -211,6 +228,26 @@ void GPS_task(void *argument)
...
+void CANBusLoad_Task(void *argument)
+{
+  /* USER CODE BEGIN CANBusLoad_Task */
+  /* Infinite loop */
+  for(;;)
+	  {
+	    CANLOAD_update_sliding_window();
+	    CAN_tx_canload_msg();
+	    osDelay(100);
+	  }
```
AarjavJain101: Fix indentation. curly brackets should be aligned with `for`

components/tel/Core/Src/freertos.c
```diff
@@ -211,6 +228,26 @@ void GPS_task(void *argument)
...
+/* USER CODE END Header_CANBusLoad_Task */
+void CANBusLoad_Task(void *argument)
+{
+  /* USER CODE BEGIN CANBusLoad_Task */
+  /* Infinite loop */
+  for(;;)
+	  {
+	    CANLOAD_update_sliding_window();
+	    CAN_tx_canload_msg();
+	    osDelay(100);
```
AarjavJain101: Make a `#define` for 100.

components/tel/Core/Src/can.c
AarjavJain101: Also in this function, you should add the bits for the canload message itself because it contributes to the CAN Bus load right. @makac1896

components/tel/Core/Inc/can.h
```diff
@@ -36,6 +36,8 @@ extern CAN_HandleTypeDef hcan;
 
 /* USER CODE BEGIN Private defines */
 #define MAX_CAN_DATA_LENGTH                 8
+#define CANLOAD_MSG_ID                      0x760
+#define CANLOAD_DATA_LENGTH                 1
```
AarjavJain101, line 40: Put both of these defines inside `can.c` since they are private.

components/tel/.cproject
AarjavJain101, line 1: Should the `.cproject` change? @kcgroulx should this file be untracked? If so, @makac1896 you can use a command like this

`git checkout origin/master -- components/tel/.cproject`

This command will pull the `.cproject` from master branch and replace your `.cproject` file with the master one.

AarjavJain101 (changes requested):

tools/CAN_Testing/BS25_CAN_ID_Table_Simulate.xmt
makac1896: Yes, that was intentional. A user can choose what load they want to simulate since this is shared across boards and no board has access to everything except TEL.

AarjavJain101: fire

kcgroulx (approved):
@makac1896 Great work!

kcgroulx (approved):

EvanO12 (approved):

EvanO12 (approved):

AarjavJain101 (approved):

PR #167, opened 2024-11-03
Feat: CAN comms moved to can.c and release build fixed

AarjavJain101:
## Pull Request Description

**Problem**
The release build of master did not work (IWDG reset occurred instantly) and also there was a coding style issue where the CAN Comms logic was implemented and initialized inside `main.c`

**Solution**
The linked monday update will explain the train of thought, however, **a temporary solution** to the release build not working has been changing **all tasks defined in `freertos.c` as statically allocated as opposed to dynamically allocated**. More research will need to be done on **what** actually changes when you change the allocation method to be static.
* According to FreeRTOS, however, doing statically allocated tasks actually takes up more memory. So its pretty suprising that this solution works.

In terms of coding style fix, i was also able to perform of CAN Comms init logic inside 1 CAN function called `CAN_tasks_init()` which you call inside `freertos.c`

## Monday Link

https://ubcsolar26.monday.com/boards/7524367653/pulses/7768018872

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DID
- [ ] ECU
- [ ] MCB
- [ ] MDI
- [x] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

Sent CAN messages using PCAN to the TEL board, observed correct messages over radio (seen on serial on my computer). LED blinking when CAN was Rxed. Confirmed that IMU, GPS, and default task were also running as well.

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


kcgroulx (approved):
Did you mean to update the can_messages.yaml?
Approved otherwise

DiegoArmstrong (approved):
( . ) Y ( . )

PR #186, opened 2025-02-07
MDI Refactor

kcgroulx:
## Pull Request Description

MDI refactor.

1. ﻿Updated the format of the motor command CAN message

- Old motor command CAN message

﻿Throttle value  (bits 0-31)  (float)
Velocity value (bits 32-64) ﻿(float)
﻿
﻿- Refactor Motor command CAN message﻿
﻿
﻿Accel DAC value (bits 0-15) (uint10)
﻿Regen ﻿DAC value (bits 16-32) (uint10)
﻿direction value (bit 33) (bool)
﻿eco mode value (bit 34) (bool)
﻿
﻿The reason why the old motor command CAN message was structured as is was to ensure that the MCB was backwards compatible with the tritium. This also made the drive system more complex as the drive state machine was calculated partially on the MCB and partially on the MDI.
﻿For the refactored CAN message, the outputs to the MDU are directly controlled via CAN. This makes the MDI as simple as possible while also making the motor command CAN message simpler to understand and work with.

2. Added timeout that will stop the MDU if no motor command CAN messages have been received in [x] ms.

﻿3. Updated to match coding standard

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DRD
- [ ] ECU
- [x] MDI
- [ ] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [x] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

- Connected to MDI with  PCAN. 
- Faked DRD CAN messages.
- Verified outputs of DACs and GPIOs using an oscilloscope.

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [x] Steps confirmed


components/mdi/Core/Src/main.c
```diff
@@ -140,189 +97,43 @@ int main(void)
...
-	}
-    /////////////////////////////////////////////////////////
-    // PARSE MESSAGES COMING FROM MOTOR CONTROLLER AND MCB //
-    /////////////////////////////////////////////////////////
-    if(Parse_Data_Flag == 1 ) //Parse_Data_Flag set in interrupt
-    {
-      CopyRxData(localRxDataMessage401,RxDataLocal);
-      CAN_Decode_Velocity_Message(RxDataLocal, &msg0); //Decode the Incoming 0x401 Messages to get data for driving the car
+  // Make sure we set DACs to 0 for redundancy.
+  MDI_stop();
```
AarjavJain101: Can we name this `MDI_stop_motor()`. Just so its clear what exactly this function is stopping.

components/mdi/Core/Src/main.c
```diff
@@ -140,189 +97,43 @@ int main(void)
...
-    if(parse_frame2 == 1) //Parse the frame based on info we want
+  while (1)
+  {
+    // If MDI hasn't received a CAN message for MAX_TIMEOUT_VALUE, stop MDI for safety
+    if(g_last_command_time + MAX_TIMEOUT_VALUE > HAL_GetTick())
     {
-    	CopyRxData(localRxDataFrame2,RxDataLocal);
-    	Decode_Frame2(RxData,&msg0);
-    	parse_frame2 = 0; //reset flag
+      MDI_stop();
```
AarjavJain101: Great! Just for safety can we add brackets around the LHS of the `>` so no compiler issue. I know there is an ordering of operations but I have been fucked by this before and its hard to debug.

components/mdi/Core/Src/main.c
```diff
@@ -140,189 +97,43 @@ int main(void)
...
-      }
-      else{ // no regen
-        Send_Voltage(0, DAC_REGEN_ADDR, &hi2c2);
-      }
-
-      //reset flag
-      send_data_flag = 0;
-    }
+      // Set motor command
+      MDI_set_motor_command(&g_MDI_motor_command);
```
AarjavJain101: Clean af!

components/mdi/Core/Src/main.c
```diff
@@ -361,343 +173,52 @@ void SystemClock_Config(void)
...
-  sClockSourceConfig.ClockPrescaler = TIM_CLOCKPRESCALER_DIV1;
-  sClockSourceConfig.ClockFilter = 15;
-  if (HAL_TIM_ConfigClockSource(&htim1, &sClockSourceConfig) != HAL_OK)
-  {
-    Error_Handler();
-  }
-  sMasterConfig.MasterOutputTrigger = TIM_TRGO_RESET;
-  sMasterConfig.MasterSlaveMode = TIM_MASTERSLAVEMODE_DISABLE;
-  if (HAL_TIMEx_MasterConfigSynchronization(&htim1, &sMasterConfig) != HAL_OK)
+  if(HAL_OK == HAL_CAN_GetRxMessage(hcan, CAN_RX_FIFO0, &CAN_Rx_header, CAN_Rx_data))
```
AarjavJain101: Can we put the callback interrupt in `can.c`

components/mdi/Core/Src/main.c
```diff
@@ -361,343 +173,52 @@ void SystemClock_Config(void)
...
-{
-
-  /* USER CODE BEGIN TIM6_Init 0 */
-
-  /* USER CODE END TIM6_Init 0 */
-
-  TIM_MasterConfigTypeDef sMasterConfig = {0};
-
-  /* USER CODE BEGIN TIM6_Init 1 */
+    if(CAN_Rx_header.IDE == DRD_MOTOR_COMMAND_CAN_ID)
```
AarjavJain101: WAIT TONY IS USING 0x400 FOR THE STR WHEEL. Lol we will talk about htis.

components/mdi/Core/Src/main.c
```diff
@@ -140,189 +97,43 @@ int main(void)
...
-    	parse_frame2 = 0; //reset flag
+      MDI_stop();
     }
 
-    //////////////////////////////////////
-    // SEND SIGNALS TO MOTOR CONTROLLER //
-    //////////////////////////////////////
-    if( (send_data_flag == 1 ) && (Parse_Data_Flag == 0) )
+    // Wait until motor command is received over CAN
+    if(g_motor_command_received == true)
```
AarjavJain101, line 126: Currently you never set this variable to false. Maybe is it to false in the if statement

components/mdi/Core/Src/mdi.c
```diff
@@ -0,0 +1,92 @@
...
+    // Truncate if voltage_value is greater than max
+    voltage_value = (voltage_value > MAX_DAC_VALUE) ? MAX_DAC_VALUE : voltage_value;
+
+    uint8_t i2c_buffer[2];
+
+    i2c_buffer[0] = voltage_value >> 8;
+    i2c_buffer[1] = voltage_value;
+
+    // TODO: Replace blocking call
+    HAL_I2C_Master_Transmit(&hi2c2, DAC_addr, i2c_buffer, sizeof(i2c_buffer), HAL_MAX_DELAY);
```
AarjavJain101, line 37: Maybe we can do with an interrupt and use the callback to check if it worked. It it didnt work then we can do an error. Lets leave that as an MMR right now.

components/mdi/Core/Src/mdi.c
```diff
@@ -0,0 +1,92 @@
...
+ * @brief Sets the DAC voltages based on the global accel and regen DAC values
+ * @note Must call MDI_parse_motor_command() to set the global accel and regen DAC values
+ */
+void MDI_set_motor_command(MDI_motor_command_t* MDI_motor_command)
+{
+    MDI_set_DAC_voltage(ACCEL_DAC, MDI_motor_command->accel_DAC_value);
+    MDI_set_DAC_voltage(REGEN_DAC, MDI_motor_command->regen_DAC_value);
+
+    HAL_GPIO_WritePin(DIR_GPIO_Port, DIR_Pin, MDI_motor_command->direction_value);
+    HAL_GPIO_WritePin(DIR_GPIO_Port, DIR_Pin, MDI_motor_command->eco_mode_value);
```
AarjavJain101: Why are these both going to the same pin? Also what is the purpose of toggling these pins?

components/mdi/Core/Src/mdi.c
```diff
@@ -0,0 +1,92 @@
...
+ * @brief Sets the inputs to the MDU to the default values for safety.
+ * Both DACs are set to 0. Direction is forward and eco mode is turned on
+ * 
+ */
+void MDI_stop()
+{
+    MDI_motor_command_t MDI_motor_command;
+    MDI_motor_command.accel_DAC_value = 0;
+    MDI_motor_command.regen_DAC_value = 0;
+    MDI_motor_command.direction_value = 1;
```
AarjavJain101: Lets add a define for direction = 1as
`#define CURRENT_TO_MOTOR      1`

components/mdi/Core/Src/can.c
```diff
@@ -1,259 +1,122 @@
...
+  hcan.Init.AutoBusOff = DISABLE;
+  hcan.Init.AutoWakeUp = DISABLE;
+  hcan.Init.AutoRetransmission = DISABLE;
+  hcan.Init.ReceiveFifoLocked = DISABLE;
+  hcan.Init.TransmitFifoPriority = DISABLE;
+  if (HAL_CAN_Init(&hcan) != HAL_OK)
+  {
+    Error_Handler();
+  }
+  /* USER CODE BEGIN CAN_Init 2 */
```
AarjavJain101, line 59: The clock config is 8Mhz. Is there a reason for this? Can we also add the high speed external clock?
image
image

AarjavJain101, line 59: Also add Rx interrupt on the `.ioc` and also add `HAL_CAN_ConfigFilter()` so you can recieve CAN messages.

components/mdi/Core/Src/can.c
```diff
@@ -1,259 +1,122 @@
...
-	CAN_msg->motorCurrentFlag = CurrentLimit;
-	CAN_msg->softwareOverCurrent = OverCurrent;
+  /* USER CODE BEGIN CAN_Init 0 */
+
+  /* USER CODE END CAN_Init 0 */
+
+  /* USER CODE BEGIN CAN_Init 1 */
+
+  /* USER CODE END CAN_Init 1 */
+  hcan.Instance = CAN1;
```
AarjavJain101, line 43: Also you will need to change the baudrate when you change the clock to 72 Mhz.

AarjavJain101 (changes requested):
good!

components/mdi/Core/Src/mdi.c
kcgroulx: whoops

components/mdi/Core/Src/main.c
kcgroulx: Changed to 0x401

components/mdi/Core/Src/mdi.c
kcgroulx: @AarjavJain101 These pins control the MDU. DIR is the direction (fwd or reverse). eco mode is the power mode (Power or Eco)

components/mdi/Core/Src/mdi.c
kcgroulx: Makes sense, I used an enum instead.

mridul2004 (approved):
looks good to me

components/mdi/Core/Src/can.c
AarjavJain101: @kcgroulx

components/mdi/Core/Src/can.c
```diff
@@ -1,259 +1,173 @@
...
-	return;
+  CAN_FilterTypeDef canFilterConfig;
+
+  canFilterConfig.FilterBank = 0;                      // Use filter bank 0 (0 to 13 for F103)
+  canFilterConfig.FilterMode = CAN_FILTERMODE_IDMASK;  // Mask mode for exact matching
+  canFilterConfig.FilterScale = CAN_FILTERSCALE_32BIT; // 32-bit filter
+  canFilterConfig.FilterIdHigh = (0x401 << 5) >> 16;   // Standard ID shifted
+  canFilterConfig.FilterIdLow = (0x401 << 5) & 0xFFFF;
+  canFilterConfig.FilterMaskIdHigh = (0x7FF << 5) >> 16; // Mask only standard ID bits
+  canFilterConfig.FilterMaskIdLow = (0x7FF << 5) & 0xFFFF;
```
AarjavJain101, line 165: Does the MDI only receive one CAN message whose ID = 0x401?

components/mdi/Core/Src/can.c
```diff
@@ -1,259 +1,173 @@
...
+  }
+  /* USER CODE BEGIN CAN_Init 2 */
 
-	//5 bit number
-	uint32_t FET_Temperature = ((uint32_t)(localRxData[4] & 0x07) << 2) |
-	 			 	           ((uint32_t)(localRxData[3] & 0xC0) >> 6) ;
+  CAN_Filter_Config();
 
-	CAN_msg -> controllerHeatsinkTemp = (float) FET_Temperature * 5.0; //datasheet 5 Celsius per LSB
+  /* USER CODE END CAN_Init 2 */
```
AarjavJain101, line 64: Can we add a CAN message for MDI that is the velocity of the car? Currently the CAN ID table says that we do 2 * pi / 60 to conver rpm to m/s. Problem with this is that we dont consider the radius/size of the wheel. We need to consider the size of the wheel to actually know the velocity of the car
image

To fix this we can scale the rpm message in the dbc file by the appropriate amount to convert from rpm to m/s of the car NOT the motor.

components/mdi/Core/Src/main.c
```diff
@@ -140,189 +97,42 @@ int main(void)
...
-    if(parse_frame2 == 1) //Parse the frame based on info we want
+  while (1)
+  {
+    // If MDI hasn't received a CAN message for MAX_TIMEOUT_VALUE, stop motor for safety
+    if((g_last_command_time + MAX_TIMEOUT_VALUE) > HAL_GetTick())
     {
-    	CopyRxData(localRxDataFrame2,RxDataLocal);
-    	Decode_Frame2(RxData,&msg0);
-    	parse_frame2 = 0; //reset flag
+      MDI_stop_motor();
```
AarjavJain101, line 122: Can we add a diagnostic message here to say that we timed out. Even though we stop the motor we nee dto kno why

AarjavJain101 (changes requested):

components/mdi/Core/Src/can.c
kcgroulx: @AarjavJain101 As of now, yes.

components/mdi/Core/Src/main.c
kcgroulx: @AarjavJain101 Yes that is the plan. Out of the scope of this PR though.

components/mdi/Core/Src/can.c
kcgroulx: @AarjavJain101 The MDU (Mitsuba) reports the motor rpm over CAN already, so I don't think it makes sense for the MDI to send it as well. The DRD can read this value and do the calculations instead.

components/mdi/Core/Src/main.c
```diff
@@ -99,7 +99,9 @@ int main(void)
   MX_I2C2_Init();
   /* USER CODE BEGIN 2 */
 
+  // Start CAN and activate Rx interrupt
   HAL_CAN_Start(&hcan);
+  HAL_CAN_ActivateNotification(&hcan, CAN_IT_RX_FIFO0_MSG_PENDING);
```
AarjavJain101: Should do HAL_CAN start after notificaiton set:
image

components/mdi/Core/Src/can.c
```diff
@@ -1,259 +1,173 @@
...
-* Takes the CAN message and decodes the data to know in which mode of operation the motor controller should be in and separates the message data into
-* the acceleration and velocity components
-*
-* @param RxData array of 8 bit wide elements that contains the CAN message sent by the SCN.
-* @param CAN_msg struct that holds the velocity, acceleration and current state of operation of the motor controller
-* @retval Modifies CAN_msg struct with the appropriate modes of operation and the values for velocity and acceleration
-*/
-void CAN_Decode_Velocity_Message(uint8_t localRxData[], CAN_message_t* CAN_msg)
-{
+/* USER CODE BEGIN 0 */
```
AarjavJain101, line 23: Add include for `mdi.h`

components/mdi/Core/Inc/mdi.h
```diff
@@ -0,0 +1,51 @@
...
+typedef enum {
+    ECO_MODE   = 0,
+    POWER_MODE = 1
+} MDI_power_mode_t;
+
+typedef struct{
+    uint16_t accel_DAC_value;
+    uint16_t regen_DAC_value;
+    bool direction_value;
+    bool eco_mode_value;
```
AarjavJain101, line 41: Can we make these volatile

components/mdi/Core/Src/can.c
```diff
@@ -1,259 +1,173 @@
...
+      // Parse motor command and update the global MDI motor command variable
+      MDI_parse_motor_command(CAN_Rx_data, &g_MDI_motor_command);
+
+      // Set motor command received flag to true
+      g_motor_command_received = true;
+    }
+  }
+  else
+  {
+    // TODO: Error handling
```
AarjavJain101, line 150: set `g_motor_command_received = false;` at the very start just in case.

components/mdi/Core/Src/can.c
AarjavJain101: @kcgroulx. Sorry my message was actually misleading. Instead of a new CAN message lets change the scale factor on the motor speed message in the DBC from `2 * pi / 60` to `2 * pi * CIRCUMFERENCE / 60`. Basically just copy the old MDI speed calculation code.

components/mdi/Core/Src/main.c
AarjavJain101: Cool. Can you add to monday in a way so that we remember.

AarjavJain101 (dismissed):
Changes

components/mdi/Core/Src/main.c
kcgroulx: Good to know, I never knew that

components/mdi/Core/Src/mdi.c
```diff
@@ -0,0 +1,92 @@
...
+ * @param voltage_value 10 bit unsigned integer. Scaled linearly to the voltage value
+ */
+void MDI_set_DAC_voltage(MDI_DAC_addr_t DAC_addr, uint16_t voltage_value)
+{
+    // Truncate if voltage_value is greater than max
+    voltage_value = (voltage_value > MAX_DAC_VALUE) ? MAX_DAC_VALUE : voltage_value;
+
+    uint8_t i2c_buffer[2];
+
+    i2c_buffer[0] = voltage_value >> 8;
```
EvanO12, line 33: Double check that this is the right order of bit shifting.

EvanO12 (approved):
lgtm

PR #188, opened 2025-02-18
Feat: STR and Turn Signal Logic

tonychen-2006:
## Pull Request Description

Added the STR file and implemented code for turn signal logic to send the status of RTS_IN or LTS_IN over CAN

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [X] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


components/str/Core/Src/main.c
```diff
@@ -0,0 +1,220 @@
...
+  /* USER CODE END 2 */
+
+  /* Infinite loop */
+  /* USER CODE BEGIN WHILE */
+  while (1)
+  {
+	  // HAL_GPIO_TogglePin(DEBUG_LED_GPIO_Port, DEBUG_LED_Pin);
+	  // HAL_Delay(500);
+
+    CAN_tx_turn_signal_msg(rts_reading, lts_reading);
```
AarjavJain101: Make these global variables start with a `g_`

components/str/Core/Inc/can.h
```diff
@@ -0,0 +1,52 @@
...
+extern CAN_HandleTypeDef hcan;
+
+/* USER CODE BEGIN Private defines */
+
+/* USER CODE END Private defines */
+
+void MX_CAN_Init(void);
+
+/* USER CODE BEGIN Prototypes */
+
```
AarjavJain101: Add your declaration for the CAN Tx turn signal function here.

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,144 @@
...
+  ******************************************************************************
+  */
+/* USER CODE END Header */
+/* Includes ------------------------------------------------------------------*/
+#include "can.h"
+
+/* USER CODE BEGIN 0 */
+
+/* PRIVATE INCLUDES */
+#define TURN_SIGNAL_CAN_DATA_LENGTH 2
```
AarjavJain101: Why are you using 2 bytes? Since the value is either 0 or 1 that takes up 1 bit. Since you have 2 variables (RTS and LTS) and each is 1 bit that takes only 2 bits. Package those 2 bits into the same 1 byte and make the CAN messgae length 1 byte.

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,144 @@
...
+
+#define TURN_SIGNAL_CAN_MESSAGE_ID  0x555 // Change
```
AarjavJain101: Looking at the CAN ID table and our overall system, what do you think the ID should be.

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,144 @@
...
+
+CAN_TxHeaderTypeDef turn_signal = {
```
AarjavJain101: name it `turn_signal_msg` to be clear.

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,144 @@
...
+void CAN_tx_turn_signal_msg(int rts_reading, int lts_reading) {
+
+    CAN_comms_Tx_msg_t CAN_comms_Tx_msg = {
+      .data[0] = (uint8_t) rts_reading,
+      .data[1] = (uint8_t) lts_reading,
+
+      .header = turn_signal
+    };
+  CAN_comms_Add_Tx_message(&CAN_comms_Tx_msg);
+  RADIO_filter_and_queue_msg_tx(&CAN_comms_Tx_msg);
```
AarjavJain101: No radio on the STR.

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,144 @@
...
+/* USER CODE BEGIN 1 */
+void CAN_tx_turn_signal_msg(int rts_reading, int lts_reading) {
+
+    CAN_comms_Tx_msg_t CAN_comms_Tx_msg = {
+      .data[0] = (uint8_t) rts_reading,
+      .data[1] = (uint8_t) lts_reading,
+
+      .header = turn_signal
+    };
+  CAN_comms_Add_Tx_message(&CAN_comms_Tx_msg);
```
AarjavJain101: No need for `CAN_comms` Just do Regular `HAL_CAN_AddTxMessage`. Remember the STR only sends 1 CAN message and receives 0 messages so no need. **Also whenever you push ANY code build it please. Right now you have not included `CAN_comms` but you are calling that function. Either way we dont need to use CAN comms anyways.**

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,144 @@
...
+
+  /* USER CODE END CAN1_MspDeInit 1 */
+  }
+}
+
+/* USER CODE BEGIN 1 */
+void CAN_tx_turn_signal_msg(int rts_reading, int lts_reading) {
+
+    CAN_comms_Tx_msg_t CAN_comms_Tx_msg = {
+      .data[0] = (uint8_t) rts_reading,
```
AarjavJain101: Since the data is only 1 byte long and bits 0 and 1 are respectively the values of the turn signals you will need to change this.

components/str/Core/Src/main.c
AarjavJain101: Right now there is no delay so this would be really bad. But we only want to transmit a CAN message when the state of ANY of the LTS or RTS changes. Otherwise dont transmit a CAN message. Do you get why @tonychen-2006 ?

AarjavJain101 (changes requested):
Make changes

components/str/Core/Src/gpio.c
```diff
@@ -0,0 +1,100 @@
...
+  ******************************************************************************
+  */
+/* USER CODE END Header */
+
+/* Includes ------------------------------------------------------------------*/
+#include "gpio.h"
+
+/* USER CODE BEGIN 0 */
+int rts_reading;
+int lts_reading;
```
AarjavJain101: Add `volatile` keyword. Search it.

components/str/Core/Src/rtc.c
AarjavJain101, line 1: Disable RTC we dont need.

components/str/.settings/language.settings.xml
kcgroulx, line 1: Delete this folder and add it to the .gitignore

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,176 @@
...
+
+  /* USER CODE END CAN1_MspDeInit 1 */
+  }
+}
+
+/* USER CODE BEGIN 1 */
+/**
+ * @brief Allows all messages to be received
+ */
+void CAN_filter_init()
```
kcgroulx: This function is never called.

We should be passing the filter by reference: `void CAN_filter_init(CAN_FilterTypeDef* can_filter)`

components/str/Core/Src/can.c
```diff
@@ -0,0 +1,176 @@
...
+    can_filter.FilterScale = CAN_FILTERSCALE_16BIT;
+    can_filter.FilterActivation = CAN_FILTER_ENABLE;
+}
+
+/**
+ * @brief CAN message for true/false of the RTS_IN and LTS_IN pins
+ * 
+ * Packages the two variables into a one byte CAN message
+ */
+void CAN_tx_turn_signal_msg(int rts_reading, int lts_reading) {
```
kcgroulx: Use `uint8_t` or `bool` instead of `int`

components/str/Core/Src/gpio.c
```diff
@@ -0,0 +1,108 @@
...
+  */
+/* USER CODE END Header */
+
+/* Includes ------------------------------------------------------------------*/
+#include "gpio.h"
+
+/* USER CODE BEGIN 0 */
+volatile int g_rts_reading = 0;
+volatile int g_lts_reading = 0;
+volatile int g_reading_status;
```
kcgroulx: Good use of volatile. Change the datatype to a `bool` or `uint8_t`.

kcgroulx (changes requested):

components/str/Core/Src/can.c
kcgroulx: How about `0x580`

components/str/Core/Src/main.c
```diff
@@ -0,0 +1,222 @@
...
+    if(g_rts_reading == HAL_GPIO_ReadPin(RTS_IN_GPIO_Port, RTS_IN_Pin) || g_lts_reading == HAL_GPIO_ReadPin(LTS_IN_GPIO_Port, LTS_IN_Pin))
+    {
+      CAN_tx_turn_signal_msg(g_rts_reading, g_lts_reading);
+
+      g_rts_reading = !(HAL_GPIO_ReadPin(RTS_IN_GPIO_Port, RTS_IN_Pin));
+
+      g_lts_reading = !(HAL_GPIO_ReadPin(LTS_IN_GPIO_Port, LTS_IN_Pin));
+    }
+
+    HAL_Delay(200);
```
AarjavJain101: Make a define for this.

AarjavJain101 (changes requested):

components/str/Core/Src/gpio.c
```diff
@@ -0,0 +1,74 @@
...
+  ******************************************************************************
+  */
+/* USER CODE END Header */
+
+/* Includes ------------------------------------------------------------------*/
+#include "gpio.h"
+
+/* USER CODE BEGIN 0 */
+volatile bool g_rts_reading = 0;
+volatile bool g_lts_reading = 0;
```
AarjavJain101: These are not used in this file. Move it to `main.c` where it is used.

kcgroulx:
Make sure to fill out the PR template completely

AarjavJain101 (approved):
Looks good!

kcgroulx (approved):
✅ Great work!

PR #189, opened 2025-02-22
[DRAFT] Feat: DRD External Lights Logic

EvanO12:
## Pull Request Description

Initial logic for managing turn signals, hazard signals, and daytime running lights.
PR just for Code review.


components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+ * @brief State Machine to handle Vehicles External lights
+ * 		Manages left and right turn signals, hazard signals, and
+ * 		daytime running lights.
+ *
+ */
+void ExternalLights_state_machine()
+{
+	uint32_t dtr = HAL_GPIO_ReadPin(DTR_IN_Pin, DTR_IN_Pin);
+	uint32_t hazard_on = HAL_GPIO_ReadPin(HAZARD_GPIO_Port, HAZARD_Pin);
+	static uint8_t prev_state = DTR_STATE;
```
AarjavJain101: Add comment to explain why you need to know previous state

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+	static uint8_t count = 0;
+	static uint8_t lts = 0;
+	static uint8_t rts = 0;
+
+	if (hazard_on)
+	{
+		if (prev_state != HAZARD_STATE)
+		{
+			count = 0;
+		}
```
AarjavJain101: I dont understand why/what this does. Add a comment to explain/make mor readable

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+
+		count++;
+
+		if (count >= 10)
+		{
+			count = 0;
+			lts = !lts;
+			rts = !rts;
+			dtr = 0;
+		}
```
AarjavJain101: So I am guessing your idea is that if the task is 50ms then waiting 10 counts is 500ms. So toggle the lights at 500ms correct? if thats the logic make it more readbale. Either add comments or change the name `count` to something else more obvious.

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+		lts = 0;
+		dtr = 0;
+		prev_state = RTS_STATE;
+	}
+	else
+	{
+		lts = 0;
+		rts = 0;
+		prev_state = DTR_STATE;
+	}
```
AarjavJain101: Be clear that this else represents the DTR state. In fact make it more readabl by making it an else if instead of an else

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		prev_state = HAZARD_STATE;
+	}
+	else if (left_turn_signal)
+	{
+		if(prev_state != LTS_STATE)
+		{
+			count = 0;
+		}
```
AarjavJain101: OK I see alot of these at the start of your if statements for the state. I get the point is to reset count IF you are just starting in that state, otherwise you want to preserve the count so that it gets added to so you hit your 500ms. Can you add a comment to explain that.

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+
+		count++;
+
+		if (count >= 10)
+		{
+			count = 0;
+			lts = !lts;
+			rts = !rts;
+			dtr = 0;
```
AarjavJain101: Shouldnt `dtr = 0` be outside the if statement. For readability this helps because then I think that str being 0 has something to do with when the lights are about to toggle. But the truth is that dtr is permanently off when we in state of toggling lts rts or both

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+		case TURN_SIGNAL_LEFT:
+			left_turn_signal = 1;
+			right_turn_signal = 0;
+			break;
+
+		case TURN_SIGNAL_RIGHT:
+			right_turn_signal = 1;
+			left_turn_signal = 0;
+			break;
+	}
```
AarjavJain101: I like the idea, but lets simplify this. I dont think we should pass in a struct dedicated to the signals. Instead here is how I imagine the code flow:

In the CAN Comms Rx callback function we have a switch statement for the different CAN IDs that the DRD needs to see. The variable to switch based off of is the CAN ID of the incoming CAN message. One of the `case`'s in the switch statement will be `TURN_SIGNAL_CAN_ID`. When you go into this case, you simply call the `void External_Lights_set_turn_signals` function you made, **however, you pass in the CAN message into the function itself**. Then inside the function you made you just parse the first 2 bits of the CAN message to be the turn signals.

@kcgroulx and @EvanO12 thoughts.

components/drd/Core/Src/external_lights.c
```diff
@@ -0,0 +1,157 @@
...
+
+/*	Symbolic Constants		*/
+#define DTR_STATE		0
+#define HAZARD_STATE 	1
+#define LTS_STATE 		2
+#define RTS_STATE 		3
+
+/*	Global Variables	*/
+volatile uint8_t left_turn_signal = 0;
+volatile uint8_t right_turn_signal = 0;
```
AarjavJain101, line 29: Good!

AarjavJain101 (changes requested):
Great work so far! Changes are small and mostly documentation

components/drd/Core/Src/external_lights.c
EvanO12: yep thats the reason of count. ok will do!

components/drd/Core/Src/external_lights.c
EvanO12: yeah it should. good catch.

components/drd/Core/Src/external_lights.c
EvanO12: that sounds good to me.

AarjavJain101 (approved):
lgtm

PR #199, opened 2025-03-07
Feat: DRD Drive State.c additions

EvanO12:
## Pull Request Description

Code Review for drive_state.c on the DRD


components/drd/Core/Src/adc.c
```diff
@@ -199,4 +199,11 @@ void HAL_ADC_MspDeInit(ADC_HandleTypeDef* adcHandle)
 
 /* USER CODE BEGIN 1 */
 
+uint16_t Read_ADC(ADC_HandleTypeDef* hadc)
+{
+	HAL_ADC_Start(hadc);
+	HAL_ADC_PollForConversion(hadc, HAL_MAX_DELAY);
+	return HAL_ADC_GetValue(hadc);
```
AarjavJain101, line 206: You will need to change the config of the hadc when reading adc1 vs adc2. This is because they are on different channels. You might also need to change the sampling rate:
image

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,349 @@
...
+static void update_input_flags();
+static void get_accel_readings();
+static uint8_t get_command_flags();
+static void clear_request_flags();
+
+
+/*	Global Variables	*/
+input_flags_t input_flags;
+drive_state_t state = PARK;
+uint16_t throttle_DAC = 0;
```
AarjavJain101: Can we add `g_` prefix to all globals.

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,349 @@
...
+/*		DRIVE STATE MACHINE HANDLING	*/
+
+/**
+ * @brief Drive State Machine. Uses drive state to determine motor command
+ * 		to be sent to the MDI.
+ */
+void Drive_State_Machine_handler()
+{
+	motor_command_t motor_command;
+	update_input_flags(); //Todo: handle regen adc input
```
AarjavJain101: What is regen adc?

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,349 @@
...
+/*
+ * @brief Normalizes adc1 and adc2 into a single accel DAC value.
+ *
+ * @param adc1 - 12 bit ADC reading of first accelerator potentiometer
+ * @param adc2 - 12 bit ADC reading of second accelerator potentiometer
+ */
+void normalize_adc_values(uint16_t adc1, uint16_t adc2)
+{
+	//Todo: confirm ADC orientation
+	uint16_t uniform_adc = ((adc1 + adc2)/2) >> 2; //get rid of lowest 2 bits
```
AarjavJain101: WHy get rid of lowest 2 bits?

AarjavJain101: 1. Why are we not using the same logic/zones as we did previously on the MCB code for this functon?
2. When we send the digital value from DRD to the MDI so that it cna be directly used for the DACs to the motor, what range of values is the DAC expecting? Is it 0-4095? Or is it 10 bit and hence you cut off lowest 2 bits?

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,349 @@
...
+	uint8_t regen_first_byte = (uint8_t) (motor_command->accel_DAC_value & 255);
+	uint8_t regen_second_byte = (uint8_t) (motor_command->accel_DAC_value >> 8);
+
+	data[0] = accel_first_byte;
+	data[1] = accel_second_byte;
+	data[2] = regen_first_byte;
+	data[3] = regen_second_byte;
+	data[4] = motor_command->motor_command_flags;
+
+	memcpy(msg.data, data, CAN_DATA_SIZE);
```
AarjavJain101, line 209: Why is this 8 bytes?

components/drd/Core/Src/can.c
```diff
@@ -23,10 +23,24 @@
...
+/**
+ * 	CAN message header for a drive command. This command header is to
+ * 	send an appropriate drive command to the motor controller.
+ */
+const CAN_TxHeaderTypeDef drive_command_header = {
+    .StdId = MOTOR_DRIVE_COMMAND_ADDRESS,
+    .ExtId = 0x0000,
+    .IDE = CAN_ID_STD,
+    .RTR = CAN_RTR_DATA,
+    .DLC = CAN_DATA_SIZE
```
AarjavJain101: Why is this 8 bytes?

components/drd/Core/Inc/drive_state.h
```diff
@@ -0,0 +1,84 @@
...
+	volatile bool throttle_ADC_out_of_range;
+	volatile bool eco_mode_on;
+	volatile uint16_t accel_DAC_value;
+	volatile uint16_t regen_DAC_value; //currently unused, but we could implemented down the line
+} input_flags_t;
+
+typedef struct {
+	uint16_t accel_DAC_value;
+	uint16_t regen_DAC_value;
+	uint8_t motor_command_flags; //direction value first bit and eco mode value second bit.
```
AarjavJain101: Should these be volatile?

components/drd/Core/Src/drive_state.c
AarjavJain101, line 47: Should this be volatile?

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,349 @@
...
+}
+
+void handle_state_transition()
+{
+	bool park_request = input_flags.park_state_request;
+	bool forward_request = input_flags.forward_state_request;
+	bool reverse_request = input_flags.reverse_state_request;
+
+	//too many requests have been triggered
+	if (park_request + forward_request + reverse_request > 1)
```
AarjavJain101: Put brakcets around the LHS.

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,349 @@
...
+
+	CAN_comms_Add_Tx_message(&msg);
+}
+
+void Drive_State_can_rx_handle(CAN_comms_Rx_msg_t* can_msg)
+{
+
+	//Todo: write helper functions for parsing these messages
+	// Velocity from MDU 0x08850225
+	//BMS Regen Disabled (BMS Status 0x622)
```
AarjavJain101: BMS regen disabled is something we can consider. Previously we never used this data. Send a message tagging me Saman kyle and krish about this on slack.

AarjavJain101 (changes requested):
Changes needed

components/drd/Core/Src/drive_state.c
EvanO12: decide regen dac based on state of charge

components/drd/Core/Src/drive_state.c
EvanO12: 12 bits from adc to 10 bits which is what motor expects

EvanO12: 1. 2 adcs is different. also format we sent it to mdu with is different. logic for deciding zones ie voltages that are too high or low should be the same as mcb. if you have a specific question can you point it out to me in thursday?
2. 10 bit 0-1023

components/drd/Core/Src/adc.c
EvanO12: i pass hadc1 or hadc2 into my read adc function so I think it should be fine? There are two adc init functions in adc.c, one for each adc. Can you clarify where I may be going wrong?

kcgroulx: @AarjavJain101 What do you mean by changing the configuration?

components/drd/Core/Src/drive_state.c
EvanO12: input_flags: no, because individual fields are volatile in the struct 
drive_state is only updated by a single task, so no
throttle is only set by a single task, so no

components/drd/Core/Inc/drive_state.h
EvanO12: no

components/drd/Core/Src/can.c
EvanO12: changed it to be 5 bytes!

components/drd/Core/Inc/can.h
```diff
@@ -35,20 +35,28 @@ extern "C" {
...
 #define CAN_ID_BATT_FAULTS		 	0x622
 #define CAN_ID_PACK_VOLTAGE		    0x623
 #define CAN_ID_MTR_FAULTS         	0x08A50225
 #define CAN_ID_MTR_DATA_REQUEST     0x08F89540 //DRD sends data request to the MCU
 #define CAN_ID_TURN_SIGNALS       	0x580 
 
+#define MDU_REQUEST_COMMAND_ID 			0x08F89540
+#define FRAME0 0x08850225
+#define STR_CAN_MSG_ID					0x580
+#define MDU_REQUEST_FRAME_2_0 5
```
AarjavJain101: Dont make this 5. Use a binary format to make it clear. add a comment explainig how this works. Ie `0b101`

components/drd/Core/Inc/can.h
```diff
@@ -35,20 +35,28 @@ extern "C" {
 extern CAN_HandleTypeDef hcan;
 
 /* USER CODE BEGIN Private defines */
+extern const CAN_TxHeaderTypeDef drive_command_header;
+extern const CAN_TxHeaderTypeDef mdu_request_header;
 
 #define CAN_ID_STD                  (0x00000000U)  /*!< Standard Id */
 #define CAN_ID_EXT                  (0x00000004U)  /*!< Extended Id */
```
AarjavJain101: Why define this here? This already exists in HAL. Can you remove it

components/drd/Core/Src/freertos.c
```diff
@@ -153,7 +170,8 @@ void StartDefaultTask(void *argument)
   /* Infinite loop */
   for(;;)
   {
-    osDelay(1);
+	Motor_Controller_query_data();
+    osDelay(100);
```
AarjavJain101: Make a define

libraries/CAN_comms/Inc/CAN_comms.h
```diff
@@ -90,6 +90,7 @@ typedef struct {
 /* Public function declarations */
 void CAN_comms_init(CAN_comms_config_t* config);
 void CAN_comms_Add_Tx_message(CAN_comms_Tx_msg_t* CAN_comms_Tx_msg);
```
AarjavJain101, line 92: Can we add an osDelay after txing a message? This is a problem on TEL right now and we are having to add in osDelay(3) after each Tx or else we dont get all our messages txed. @kcgroulx

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+
+		case PARK:
+			motor_command = park_state_handle();
+		break;
+
+		default:
+			motor_command = get_motor_command(ACCEL_DAC_OFF, REGEN_DAC_OFF);
+	}
+
+	motor_command_package_and_send(&motor_command, false);
```
AarjavJain101, line 80: Make a define for false.

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+ *
+ * @param motor_command, struct containing all motor command information to be sent over CAN
+ */
+void motor_command_package_and_send(motor_command_t* motor_command, bool from_ISR)
+{
+	CAN_comms_Tx_msg_t msg;
+	msg.header = drive_command_header;
+	uint8_t data[8] = {0};
+
+	uint8_t accel_first_byte = (uint8_t) (motor_command->accel_DAC_value & 255);
```
AarjavJain101, line 198: Dont use 255. use the hex or binary represetnation for readability

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+ */
+void Drive_State_can_rx_handle(uint32_t msg_id, uint8_t* data)
+{
+	switch(msg_id)
+	{
+		case(FRAME0):
+
+			uint32_t rpm = (data[4] >> 3) | ((data[5] & 0x7f) << 5); //35th to 46th bit
+			float velocity = (WHEEL_RADIUS * 2.0 * M_PI * rpm) / 60.0;
+			velocity_kmh = velocity * 3.6;
```
AarjavJain101, line 234: Make define

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+ * @brief Handles Received CAN Messages relevant
+ */
+void Drive_State_can_rx_handle(uint32_t msg_id, uint8_t* data)
+{
+	switch(msg_id)
+	{
+		case(FRAME0):
+
+			uint32_t rpm = (data[4] >> 3) | ((data[5] & 0x7f) << 5); //35th to 46th bit
+			float velocity = (WHEEL_RADIUS * 2.0 * M_PI * rpm) / 60.0;
```
AarjavJain101, line 233: Make a define for rpm -> car_velocity

components/drd/Core/Inc/drive_state.h
```diff
@@ -0,0 +1,87 @@
...
+	CRUISE = (uint8_t) 0x02,
+	PARK = (uint8_t) 0x03,
+	REVERSE = (uint8_t) 0x04
+} drive_state_t;
+
+/*	Global Variables	*/
+
+extern input_flags_t input_flags;
+extern volatile drive_state_t drive_state;
+extern volatile bool eco_mode;
```
AarjavJain101, line 77: Whats the point of this variable? I dont see it used anywhere.

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+static void brake_press_handle();
+static void update_input_flags();
+static void get_accel_readings();
+static uint8_t get_command_flags();
+static void clear_request_flags();
+
+
+/*	Global Variables	*/
+volatile uint32_t velocity_kmh = 0;
+volatile bool eco_mode = true;
```
AarjavJain101, line 44: We dont use this? remove

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+			else
+			{
+				input_flags.velocity_under_threshold = false;
+			}
+
+		break;
+
+		case(STR_CAN_MSG_ID):
+			input_flags.eco_mode_on = (data[0] >> 2); //third bit of steering CAN message
+			eco_mode = input_flags.eco_mode_on; //global variable for LCD
```
AarjavJain101, line 249: We dont use `eco_mode` so we should remove it.

components/drd/Core/Src/drive_state.c
```diff
@@ -0,0 +1,457 @@
...
+			else if (value == 1)
+			{
+				input_flags.reverse_state_request = true;
+			}
+			else if (value == 2)
+			{
+				input_flags.forward_state_request = true;
+			}
+		break;
+	}
```
AarjavJain101, line 267: Reminder to remove this later.

components/drd/Core/Inc/drive_state.h
```diff
@@ -0,0 +1,87 @@
...
+
+#define REGEN_DAC_ON 1023
+#define REGEN_DAC_OFF 0
+#define ACCEL_DAC_OFF 0
+
+#define ADC_NO_THROTTLE_MAX 1000 //figure out
+#define ADC_FULL_THROTTLE_MIN 1850 //figure out
+#define ADC_THROTTLE_MAX	0 //figure out
+#define ADC_LOWER_DEADZONE 900 //tbd
+#define ADC_UPPER_DEADZONE 4000 //tbd
```
AarjavJain101, line 40: Reminder to not forget to change this

AarjavJain101 (changes requested):
Look sreally good. Small changes

components/drd/Core/Src/adc.c
AarjavJain101: Yea nvm I misread the arguement you passed in.

AarjavJain101 (approved):
lgtm

PR #202, opened 2025-03-12
Feat: GPS and IMU Reliability increased. Code Quality++

AarjavJain101:
## Pull Request Description

GPS was not getting parsed correctly someitmes and it was because the I2C read would happen non-blocking. That means checking if the gps_read_okay right after means its obviously not going to e ok since we JUST started I2C. Then we go back to the start of the loop and to I2C read again. The process repeat. Basically we need an osDelay or some delay to occur before checking if gps_read_okay is true. Normally you get that delay by nature of a callback telling you you are done, however, in this case, the I2C done callback just sets the gps_read_okay variable BUT we dont act on it in time in the loop :(. 

## Monday Link

https://ubcsolar26.monday.com/boards/7524367653/pulses/7524368309/posts/3936099916

## Effected Components

- [ ] AMB
- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [ ] STR
- [X] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


kcgroulx:
@AarjavJain101 I see STR changes, did you branch off master?

PR #229, opened 2025-03-29
Feat: STR Watchdog CAN message

tonychen-2006:
## Pull Request Description

Implemented a watchdog feature that is sent over CAN on the STR board.

## Effected Components

- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [X] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [X] Steps confirmed


components/str/Core/Src/iwdg.c
```diff
@@ -92,10 +92,10 @@ bool IWDG_is_reset()
...
 {
-  // TODO: Add diagnostic logic
-
   if (IWDG_is_reset())
   {
+    g_str_diagnostic_flags.bits.str_crash_iwdg = true;
+
     for (int i = 0; i < 2; i++)
     {
       HAL_GPIO_TogglePin(DEBUG_LED_GPIO_Port, DEBUG_LED_Pin);
```
kcgroulx: Let's decide if we need to blink the LED if there is a watchdog reset.

components/str/Core/Src/iwdg.c
```diff
@@ -92,10 +92,10 @@ bool IWDG_is_reset()
  */
```
kcgroulx, line 92: Make sure the comments are accurate to STR. There's a few times TEL is used instead.

kcgroulx (changes requested):

PR #236, opened 2025-04-05
Feat: Rename ECU FSM states 

CKalitin:
## Pull Request Description
Since FSGP the boards on the car have changed, this pr renames some of the boards and removes those that are no longer in use.

Rename MCB -> DRD
Remove DASH
Remove AMB

States before pr:
TELEM_ON,
MEM_ON,
DASH_ON,
MCB_ON,
MDU_ON,
AMB_ON

States after:
TELEM_ON,
MEM_ON,
DRD_ON,
MDU_ON

## Monday Link
[]
https://ubcsolar26.monday.com/boards/7524367629/pulses/8803330334/posts/3989762265

## Effected Components

- [ ] BMS
- [ ] DRD
- [X] ECU
- [ ] MDI
- [ ] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [X] Vehicle testing
- [X] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [X] Steps confirmed


AarjavJain101:
Wait when did you test this in car?

krishd12:
@CKalitin I see you didn't do two things.
1. You didn't rename the MCB_CTRL_GPIO_PORT and MCB_CTRL_PIN to be replace MCB with DID.
2. Also changing the pin labels in the IOC file of the ECU to account for the naming change in pt 1.

Can you please commit these changes aswell?

I dont believe anything should be changed, what do you think @AarjavJain101 @niksiar ?

components/ecu/ecu_firmware/Core/Inc/fsm.h
```diff
@@ -87,10 +85,8 @@ void LLIM_closed();
 void check_HLIM();
 void TELEM_on();
 void MEM_on();
-void DASH_on();
-void MCB_on();
+void DRD_on();
 void MDU_on();
```
AarjavJain101, line 89: We should rename MDU to MDI no? Is the battery FSM controlling the Motor Controller? Or is it the MDI (PCB by EMD)?

components/ecu/ecu_firmware/Core/Inc/main.h
```diff
@@ -97,12 +97,12 @@ void Error_Handler(void);
...
-#define DID_CTRL_Pin GPIO_PIN_12
-#define DID_CTRL_GPIO_Port GPIOB
-#define AMB_CTRL_Pin GPIO_PIN_13
-#define AMB_CTRL_GPIO_Port GPIOB
-#define MCB_CTRL_Pin GPIO_PIN_14
-#define MCB_CTRL_GPIO_Port GPIOB
+#define Unused1_CTRL_Pin GPIO_PIN_12
+#define Unused1_CTRL_GPIO_Port GPIOB
+#define Unused2_CTRL_Pin GPIO_PIN_13
+#define Unused2_CTRL_GPIO_Port GPIOB
```
AarjavJain101, line 103: Why do you have a bunch of unused pins? You should remove these **if they are actually unsued**.

components/ecu/ecu_firmware/Core/Src/fsm.c
```diff
@@ -410,43 +410,23 @@ void MEM_on()
...
+ * Exit Action: Set DRD pin high.
  * Exit State: MDU_ON
  */
-void MCB_on()
+void DRD_on()
 {
     if (timer_check(LVS_INTERVAL, &(ticks.last_generic_tick) ))
     {
-        HAL_GPIO_WritePin(MCB_CTRL_GPIO_Port, MCB_CTRL_Pin, HIGH);
+        // Between revision of the car MCB was combined into the DRD.
```
AarjavJain101: You dont need this comment because MCB doesnt exist in FW anymore. This is something documented on the Wiki rather than FW

AarjavJain101 (commented):
Looks good minor changes

components/ecu/ecu_firmware/Core/Inc/fsm.h
shinsamuel: This is a little confusing for me too, but I think what it's controlling is MDI, so should be renamed for convenience.

components/ecu/ecu_firmware/Core/Inc/main.h
shinsamuel: I guess this was written in case we needed to use these pins..? Let me check the cube ide but should be unused.

components/ecu/ecu_firmware/Core/Inc/fsm.h
shinsamuel: Screenshot 2025-04-08 131948

components/ecu/ecu_firmware/Core/Src/fsm.c
CKalitin: Wrote this before renaming the IOC pin from MCB to DRD, deleting comment makes sense now

components/ecu/ecu_firmware/Core/Inc/fsm.h
CKalitin: On the ECU Altium it's called MDU, on Dist Altium it's called MDU, then on the wiring harness it's called MDI.

Some sort of coordination was not done here, might be worth coming up with completely new acronyms from the start for v4 to ensure this doesn't happen again.

At this point renaming it would require changing BMS and ECU firmware and several Altium schematics. Might be worth saving this change for v4.

components/ecu/ecu_firmware/Core/Inc/main.h
CKalitin: It's an unused control pin. Meaning it's meant to toggle power to a board/spare connector but isn't wired anywhere on the distribution board.

This should be documented somewhere, maybe better in a Monday Update but I feel it's best to make it clear in the IOC that these are supposed to be use as control pins.

Thoughts?

PR #244, opened 2025-05-03
Feat: Added the horn enable pin and CAN message

tonychen-2006:
## Pull Request Description

Added the horn enable pin and sending the status over CAN.

## Effected Components

- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [X] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


components/str/Core/Inc/main.h
```diff
@@ -81,6 +83,12 @@ typedef enum {
   POWER_MODE = 0b00,
   ECO_MODE = 0b01,
 } mode_status_t;
+
+typedef enum {
+  HORN_ON = 0b00,
+  HORN_OFF = 0b01,
```
AarjavJain101: Why is HORN OFF 1? On CAN it would be more human readable to have HORN_ON as 1.

components/str/Core/Src/gpio.c
```diff
@@ -71,6 +71,12 @@ void MX_GPIO_Init(void)
   GPIO_InitStruct.Pull = GPIO_NOPULL;
   HAL_GPIO_Init(MODE_PW_GPIO_Port, &GPIO_InitStruct);
 
+  /*Configure GPIO pin : PtPin */
+  GPIO_InitStruct.Pin = HORN_EN_Pin;
+  GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
+  GPIO_InitStruct.Pull = GPIO_NOPULL;
```
AarjavJain101, line 77: Ask PAS whether the HORN EN GPIO is active low or high.

AarjavJain101 (changes requested):
Need some fixes

components/str/Core/Src/gpio.c
AarjavJain101: We know its Active Low so make this GPIO Pull up to match that. This means that you wont get much noise on the line since its being pulled up.

AarjavJain101 (changes requested):

AarjavJain101 (approved):
LGTM tested.

PR #261, opened 2025-06-07
Feat: Capped by 50% < 99V

AarjavJain101:
## Effected Components

- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [ ] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


EvanO12 (commented):
Would an algorithm that limits throttle at 50% instead of cutting every throttle by half be better? It would preserve the pedal curve for the driver

PR #267, opened 2025-08-02
Feat: Fixing TEL Crashing and Incorporating Rate Configuration in GPS

tonychen-2006:
## Pull Request Description

Fixing up GPS code in nmea_parse.c

## Effected Components

- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [ ] STR
- [X] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [X] Steps confirmed


components/tel/Core/Src/gps.c
```diff
@@ -11,15 +11,31 @@
 #include "CAN_comms.h"
 #include "radio.h"
 #include "bitops.h"
+#include "stm32f1xx_hal.h"
```
AarjavJain101: You shouldnt do this. Just include `main.h`. @EvanO12

components/tel/Core/Src/gps.c
```diff
@@ -377,3 +393,70 @@ void CAN_tx_gps_data_msg(GPS* gps_data) {
...
+}
+
+/**
+ * @brief GPS measurement rate configuration message sent over I2C
+ * 
+ * This function builds a UBX CFG-VALSET message that sets the receiver’s 
+ * measurement period (CFG-RATE-MEAS) to MEAS_MS in a selected memory layer. 
+ * It assembles the frame, computes the UBX checksum, then sends it over I2C.
+ * 
+ * @return Sends the configured MEAS_MS over I2C
```
AarjavJain101, line 426: @tonychen-2006 You also need to document exactly how these functions work on the Embedded wiki page under GPS (make a page if it doesnt exist. However, It should since you made it last year). What I mean is documenting what UBX is, what manual did you use to help you make these functions, walk through exactly your thought process for how to write to the GPS. This is crucial for our next generation to learn how to use our peripherals properly like you are!

components/tel/Core/Src/gps.c
```diff
@@ -377,3 +393,70 @@ void CAN_tx_gps_data_msg(GPS* gps_data) {
...
+    memcpy(&frame[6], payload, len);
+
+    uint8_t ckA;
+    uint8_t ckB;
+
+    ubx_cksum(&frame[2], 4 + len, &ckA, &ckB);
+
+    frame[6 + len] = ckA;
+    frame[7 + len] = ckB;
+
```
AarjavJain101: Really great work with the check sum calculation Tony!

components/tel/Core/Src/gps.c
```diff
@@ -377,3 +393,70 @@ void CAN_tx_gps_data_msg(GPS* gps_data) {
...
+    frame[3] = UBX_ID_VALSET;
+    frame[4] = (uint8_t)(len & 0xFF);
+    frame[5] = (uint8_t)((len >> 8) & 0xFF);
+    
+    memcpy(&frame[6], payload, len);
+
+    uint8_t ckA;
+    uint8_t ckB;
+
+    ubx_cksum(&frame[2], 4 + len, &ckA, &ckB);
```
AarjavJain101: Why is it 4 + len? Can you add a comment or use a define to explain what 4 is.

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,56 @@
...
+ * of data.
+ *
+ * @param sentence The input string to split
+ * @param tokens The array to store the split tokens
+ * @param max_value The maximum number of tokens to extract
+ */
+void split_commas(char *sentence, char *tokens[], int max_value) {
+    int count = 0;
+    char *pos = sentence;
+    while (pos && count < max_value) {
```
AarjavJain101: Put brackets around `count < max_value`

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,56 @@
...
+        char *comma = strchr(pos, ',');
+        if (comma) *comma = '\0';
```
AarjavJain101: Are there any problems with replace the comma with a `\0`. Can you list out any and all issues so we are aware of edge cases that can go wrong?

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,56 @@
...
+        tokens[count++] = pos;
+        pos = comma ? (comma + 1) : NULL;
```
AarjavJain101: What is this doing? Could you add a comment?

components/tel/Core/Src/nmea_parse.c
```diff
@@ -59,45 +105,49 @@ int gps_checksum(char *nmea_data)
...
 
-        char lon_d[3];
-        char lon_m[7];
+        char lon_d[4];
+        char lon_m[8];
 
-        for (int z = 0; z < 3; z++) lon_d[z] = values[4][z]; // Extract longitude degrees
-        for (int z = 0; z < 6; z++) lon_m[z] = values[4][z + 3]; // Extract longitude minutes
+        for (int z = 0; z < 3 && values[4] && values[4][z]; z++) lon_d[z] = values[4][z]; 
+        lon_d[3] = '\0';
```
AarjavJain101, line 174: Definitely need to document all of the data extraction. Explain each line and the overall goal using example sentences and show how the sentence is extracted at each line. Lets make our EMD documentation great!

components/tel/Core/Src/nmea_parse.c
```diff
@@ -201,47 +249,49 @@ int nmea_GPGSA(GPS *gps_data, char* inputString)
...
-        char lon_m[7];
-        char lonSide = values[4][0];
-        for (int z = 0; z < 3; z++) lon_d[z] = values[3][z]; // Extract longitude degrees
-        for (int z = 0; z < 6; z++) lon_m[z] = values[3][z + 3]; // Extract longitude minutes
+        char lon_d[4];
+        char lon_m[8];
+        char lonSide = values[4] ? values[4][0] : '\0';
+        for (int z = 0; z < 3 && values[3] && values[3][z]; z++) lon_d[z] = values[3][z];
+        lon_d[3] = '\0';
+        for (int z = 0; z < 6 && values[3] && values[3][z + 3]; z++) lon_m[z] = values[3][z + 3];
```
AarjavJain101, line 293: Similarly here make sure to document everything!

components/tel/Core/Src/nmea_parse.c
```diff
@@ -333,32 +355,27 @@ int nmea_GPRMC(GPS *gps_data, char* inputString)
  */
 int nmea_GPVTG(GPS *gps_data, char* inputString) 
 {
-    char *values[25];
-    int counter = 0;
+    char sentence[NMEA_MAX_SENTENCE];
+    // strncpy(line, inputString, sizeof(line)-1);
+    // line[sizeof(line)-1] = '\0';
+
```
AarjavJain101: One thing that was confusing to me is that _multiple_ GPS message types (GPVTG and GPGLL and more) all report long and lat. Is that true? Can you document on the wiki what all the messages are and which have long and lat. Also documenting _why_ we extract long and lat from all of these messages. Why not just extract from GPGLL?

components/tel/Core/Src/nmea_parse.c
```diff
@@ -462,8 +467,4 @@ void nmea_parse(GPS *gps_data, uint8_t *buffer) {
             }
         }
     }
-    for(int i = 0; i<cnt; i++) 
-    {
-        if (data[i] != NULL) { free(data[i]); }
-    }
-}
\ No newline at end of file
+}
```
AarjavJain101, line 477: @tonychen-2006 Can you comment out any code that is not extracting long and lat. We should **only** be extracting long, lat, Fixqulaity, number of satellites. Ex. SNR is not needed. This way we reduce computation needed and make GPS parsing simpler. We only actually need long and lat for STG.

AarjavJain101 (changes requested):
Great work so far! Lets focus on documentation now and wrap up GPS!

components/tel/Core/Src/nmea_parse.c
EvanO12: We can also configure the receiver to not send these nmea messages, although that is more work for not much benefit

components/tel/Core/Src/freertos.c
```diff
@@ -273,8 +275,10 @@ void GPS_task(void *argument)
 {
   /* USER CODE BEGIN GPS_task */
     /* Infinite loop */
-    
+
     osDelay(GPS_TASK_OFFSET_DELAY);
+    osDelay(5000);
```
EvanO12: why is there an additional 5 second delay here?

components/tel/Core/Inc/gps.h
```diff
@@ -32,6 +32,8 @@ void gps_task();
 
 void read_i2c_gps_module(uint8_t* receive_buffer);
 
+HAL_StatusTypeDef gps_config_meas_rate();
```
EvanO12, line 37: Unless we decide to check the return value of this function, we should make the return type void.

components/tel/Core/Src/freertos.c
```diff
@@ -273,8 +275,10 @@ void GPS_task(void *argument)
 {
   /* USER CODE BEGIN GPS_task */
     /* Infinite loop */
-    
+
     osDelay(GPS_TASK_OFFSET_DELAY);
+    osDelay(5000);
+    gps_config_meas_rate();   // GPS configuration to change data refresh rate
```
EvanO12: @AarjavJain101 @tonychen-2006  What are your thoughts on checking to make sure that the configuration was successful? Ie if the I2C hal status return is HAL_OK, then read the ack from the receiver. If the ack is negative, then send the config again, if the ack is positive, continue to the main for loop.

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,56 @@
...
 
+#define NMEA_MAX_SENTENCE 140
+#define GPGGA_MAX_VALUES  12
+#define GPGSA_MAX_VALUES  18
+#define GPGLL_MAX_VALUES  5
+#define GPRMC_MAX_VALUES  10
+#define GPVTG_MAX_VALUES  8
+#define GPGSV_MAX_VALUES  8
+
 char *data[20];
```
EvanO12: As per our style guide, please declare pointers as char* name!

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,56 @@
...
+ * string function 'strtok' which skips over these fields resulting in the loss
+ * of data.
+ *
+ * @param sentence The input string to split
+ * @param tokens The array to store the split tokens
+ * @param max_value The maximum number of tokens to extract
+ */
+void split_commas(char *sentence, char *tokens[], int max_value) {
+    int count = 0;
+    char *pos = sentence;
```
EvanO12: pos is not a super descriptive name. Just using the name position is fine, or maybe "token_start" or something along those lines

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,56 @@
...
+        char *comma = strchr(pos, ',');
+        if (comma) *comma = '\0';
+        tokens[count++] = pos;
+        pos = comma ? (comma + 1) : NULL;
+    }
+    if (count < max_value) tokens[count] = NULL;
+}
+
+
+void copy_sentence(char *output, const char *source, size_t size) {
```
EvanO12, line 58: Can you write a function prototype/description for this function?

EvanO12: Doesn't strncpy implement the same behaviour?

components/tel/Core/Src/nmea_parse.c
```diff
@@ -59,45 +105,49 @@ int gps_checksum(char *nmea_data)
...
     memset(values, 0, sizeof(values));
-    char *marker = strtok(inputString, ",");  // Tokenize the input string using ',' as a delimiter
 
-    while (marker != NULL) 
-    {
-        values[counter++] = malloc(strlen(marker) + 1);
-        strcpy(values[counter - 1], marker);
-        marker = strtok(NULL, ",");
-    }
+    split_commas(sentence, values, GPGGA_MAX_VALUES);
```
EvanO12, line 127: Good way of making sure we aren't doing any extraneous parsing!

components/tel/Core/Src/nmea_parse.c
```diff
@@ -417,28 +428,22 @@ void nmea_parse(GPS *gps_data, uint8_t *buffer) {
     
     memset(data, 0, sizeof(data)); // Clear array
 
-
-    char *token = strtok((char *)buffer, "$"); // Tokenize the buffer using '$' as a delimiter
-    token = strtok(NULL, "$");
-
     int cnt = 0;
```
EvanO12: Can we use a more descriptive name for this, like token_count or something along those lines?

components/tel/Core/Src/nmea_parse.c
```diff
@@ -417,28 +428,22 @@ void nmea_parse(GPS *gps_data, uint8_t *buffer) {
...
+    char *token = strtok((char *)buffer, "$");
 
-    while(token !=NULL)
-    {
-
-        data[cnt++] = malloc(strlen(token) + 1 ); // Allocate memory for each sentence
-        strcpy(data[cnt-1], token); // Copy the token into allocated memory
-        token = strtok(NULL, "$"); // Get the next sentence
+    while (token && cnt < 20) {
```
EvanO12: Can we use a symbolic constant instead of just having the magic number 20

EvanO12 (changes requested):
Good work Tony! Just some minor changes and we should be able to merge soon. As a reminder, we need to document GPS to finish up the project. 

Currently, we have a short GPS section under the TEL board in the Solar wiki. It we be good to expand on what's already there by:
1. Documenting all the GPS configuration work you did. Explain how it works, and where you got your information for it.
2. More thoroughly document nmea_parse.c. Explain how nmea_parse works(), how strings are tokenized, how we store the data, etc etc

components/tel/Core/Src/nmea_parse.c
tonychen-2006: Strncpy is very similar; however, this function **ensures** that no other line is being parsed beyond the "\r\n" newline.

components/tel/Core/Src/freertos.c
tonychen-2006: This was added during testing as we realized that cutting the power would no longer keep the configuration.

EvanO12: hmm ok. I am a little confused how the 5s delay solves this issue. Can you elaborate?

tonychen-2006: I'm not entirely sure, as it was a feature Aarjav told me to add, but it resolves this issue after running a couple of tests. Entering the delay likely gives the GPS more time to commit the configuration before continuing. A shorter delay would likely work as well.

components/tel/Core/Src/freertos.c
AarjavJain101: Yea this is a good idea. Then we should also add this status to the diagnostic bits for TEL. This way if the GPS bugs out we will know exactly why. @tonychen-2006 can you implement this?

tonychen-2006: Are we continuing only if the config is successful or just have a status that tells us whether it succeed or failed.

EvanO12: I don’t think we need a status to tell us if it failed because we can see that on sunlink, we are only getting long/lat 1/4 of the time

components/tel/Core/Src/freertos.c
EvanO12: The delay happens before we send the config message so that doesn't make a whole lot of sense to me. One possibility is that perhaps the power to the board is more volatile when we are flashing it/on startup because we might be moving the stlink around or what not. @AarjavJain101 What was your rationale behind this?

components/tel/Core/Src/freertos.c
```diff
@@ -273,8 +277,19 @@ void GPS_task(void *argument)
 {
   /* USER CODE BEGIN GPS_task */
     /* Infinite loop */
-    
+
     osDelay(GPS_TASK_OFFSET_DELAY);
+    osDelay(5000);
+
+    HAL_StatusTypeDef status = gps_config_meas_rate();
```
EvanO12, line 280: The HAL_status will not tell us if the GPS config was actually successful - it will return HAL_Ok as long as the I2C messages are sent. I was imaging that we would read the ack/nack sent by the receiver, which contains information on whether our config was saved succesfully. This is more work, but will better provide us insight into whether the GPS config has been succesful.
There's also the possibility that we lose power midway through, and the config on the GPS is reset. Did your research into the receiver's flash memory tell you anything @tonychen-2006? I remember reading that the flash layer might only be available if there is external flash memory, but I wanted to confirm that.

components/tel/Core/Src/nmea_parse.c
```diff
@@ -11,10 +11,68 @@
...
+ * 
+ * This function ensures additional sentences are not parsed by copying
+ * until the line "\r\n" is reached or the specified size if the line
+ * ending is not found.
+ *
+ * @param output The destination buffer for the copied string
+ * @param source The source string to copy from
+ * @param size The maximum number of bytes to copy
+ */
+void copy_sentence(char *output, const char *source, size_t size) {
```
EvanO12, line 58: can you change the pointer syntax to char* in the function declaration?

EvanO12: Please do this for all uses of pointers in nmea_parse.c

components/tel/Core/Src/freertos.c
EvanO12: We will leave this as an mmr for now, that we can implement later if we deem fit (based on observations from October driving day)

EvanO12 (approved):

AarjavJain101 (approved):

PR #273, opened 2026-01-28
User/cdella778/bms/fsm

AarjavJain101:
## Effected Components

- [ ] BMS
- [ ] DRD
- [ ] ECU
- [ ] MDI
- [ ] STR
- [ ] TEL
- [ ] CI/CD

## Testing

- [ ] Vehicle testing
- [ ] Bench top testing
- [ ] Unit testing
- [ ] Other
- [ ] N/A

## Sanity check

- CAN ID table updated
- IOC updated and commited
- [ ] Steps confirmed


components/ecu/ecu_firmware/Core/Inc/can.h
```diff
@@ -15,6 +15,8 @@
 /*============================================================================*/
 /* DEFINITIONS */
 
+// hheeee haaaa haaa
```
AarjavJain101, line 18: bruh

components/ecu/ecu_firmware/Core/Src/can.c
```diff
@@ -176,6 +179,11 @@ void CAN_CheckRxMessages(uint32_t rx_fifo)
             {
                 CAN_data.new_charger_msg_received = true;
             }
+            // If data is from pack message 0x623, store it for later use as uint16_t
+            if (CAN_data.rx_message.rx_header.StdId == 0x623U)
+            {
+                packVoltage = 1000*(uint16_t)(CAN_data.rx_message.data[0]) | ((uint16_t)(CAN_data.rx_message.data[1]) << 8); // store pack voltage for later use -- scaled to mV
```
AarjavJain101, line 185: packVoltage is a `uint16_t`. Did you do a calculation to **ensure**  that packVoltage will not **overflow** in this calculation for all expected CAN_data.rx_message.data?

components/ecu/ecu_firmware/Core/Src/fsm.c
```diff
@@ -303,14 +304,62 @@ void check_LLIM()
...
+    */
+    HAL_Delay(90);
+    CAN_CheckRxMessages(CAN_RX_FIFO0);
+
+    if (timer_check(BMS_STARTUP_INTERVAL, & (ticks.last_generic_tick) ))
+    {
+        FSM_state = FAULT;
+    }
+     // Fills packVolage with the latest value from the CAN message 0x623, which is pack voltage
+    else if (ecu_data.adc_data.ADC_MCPC_voltage>packVoltage) //133V in mV
```
AarjavJain101, line 344: Why will it be greater than packVoltage?

components/ecu/ecu_firmware/Core/Src/can.c
Cdella778: Did not do this -- however, it will overflow by the looks of it. 134 * 1000 = 134000, which is greater than the uint16_t max value of  2^16 or 65,535. The value here should either be stored as a uint32 or not perform the mV conversion. The ADC reading would also run into this issue, so changing them both to uint32 makes the most sense to me.

components/ecu/ecu_firmware/Core/Src/fsm.c
Cdella778: It won't be, the reason I had it setup like that was because we had been getting  000000 readings for both the ADC value and pack voltage immediately after starting, meaning a success was triggering from garbage values. Once we figure out the reading issues, I'll change the condition to being within 99% of packVoltage, so about +-1500mV.

